PR #7, opened 2014-05-09
Add a priority parameter to ensure correct loading order

conradz:
This adds a `priority` parameter to the query that allows you to specify the priority of a stylesheet.

My use case for this is that I have a "theme" stylesheet that must be applied after the default styles in the app. The theme stylesheet must be defined in `head` after all the default styles, to allow theme styles to overwrite default styles. Due to dynamic loading, sometimes the default styles are `require`d after the theme is.

This change lets you specify a priority for a stylesheet. A stylesheet with a higher priority will always be loaded into `head` after all lower-priority stylesheets. The default priority is 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,16 @@ require("style!raw!./file.css");
 
 It's recommended to combine it with the [`css-loader`](https://github.com/webpack/css-loader): `require("style!css!./file.css")`.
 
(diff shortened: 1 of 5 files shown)


sokra:
I won't merge this, because order should be defined by dependencies. Add an `@import` if you want a stylesheet after another.

PR #40, opened 2015-02-03
Support IE8

kui:
* Avoid `Function.prototype.bind`
* Force single-tag mode on IE8

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -10,8 +10,8 @@ var stylesInDom = {},
 			return memo;
 		};
 	},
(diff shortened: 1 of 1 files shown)


sokra:
> Avoid Function.prototype.bind

no, polyfills are the job of the application.

> Force single-tag mode on IE8

yes

Could you please change the PR to only include the single-tag mode fix?

kui:
I never would have send this request if I used the `bind` pollyfill. I think it can not be said _IE8 supported_ to require the polyfill.

PR #43, opened 2015-02-27
Apply default singleton to IE6-8 as well

mako-taco:
Might as well apply this fix to older browsers for those of us trying to use webpack in the horrifying world of IE8


sokra:
Thanks

PR #46, opened 2015-03-21
[HMR] fixed useable for styles with multiple refs

bloodyowl:
the issue was that the `!(refs++)` expression in `exports.ref`never hit `true` with HMR when you've got `refs` > 1.

this commit patches it without altering the behaviour in non-HMR conditions.


MoOx:
Oh nice! This should fix style hot update in react components when you use/unuse styles in components update hooks.

sokra:
Could you please check if this commit (5e5c3a0) also fixes the bug?

bloodyowl:
seems that it does. thanks!
can you publish a npm release with that fix?

sokra:
done

MoOx:
:+1:

PR #51, opened 2015-04-14
[#39] Don't use styleSheet to store state of css text

### LINKED ISSUE #39: singleton mode problems in IE9
```cssText``` in IE does not supports media queries, i.e. does not return them (I think because of backwards compatibility). So when in single mode (which is by default) you use media queries and that stylesheet with media queries was ```require```d before some other -- then you loose your media queries.

Probably for IE you should store stylesheet in memory (i.e. object/array) and when new modification/addition comes just concatenate all entries and again apply **original** string to stylesheet.

For my case I just disabled singleton mode, but it's not obvious at first glance that you need to do it.

diurnalist:
IE9 will actually drop certain things (like media queries) from the
cssText property when it is read back out. To fix, store the raw css
text in a separately managed array, and use that to form the entire
string that should be set on the style element's cssText property.

Fixes #39.

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -153,25 +153,20 @@ function addStyle(obj, options) {
 	};
 }
 
-function replaceText(source, id, replacement) {
-	var boundaries = ["/** >>" + id + " **/", "/** " + id + "<< **/"];
-	var start = source.lastIndexOf(boundaries[0]);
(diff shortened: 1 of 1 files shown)


sokra:
Thanks. It's also much cleaner and simpler... ;)

PR #59, opened 2015-04-30
Put @font-face in own style elements in IE8 and below

jerrysu:
This fixes https://github.com/webpack/style-loader/issues/58, where the use of `@font-face` was causing IE8 to render a blank page of invisible elements when subsequent changes were made to the same singleton `<style>` tag.

To fix this issue, individual `<style>` tags are created for the `@font-face` instances in a stylesheet. Yes, it is possible to hit the stylesheet limit if you use a lot of custom fonts. However, I don't think people use more than a few fonts at a time.

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -10,14 +10,17 @@ var stylesInDom = {},
 			return memo;
 		};
 	},
-	isOldIE = memoize(function() {
-		return /msie [6-9]\b/.test(window.navigator.userAgent.toLowerCase());
+	isIE = memoize(function() {
+		var match = window.navigator.userAgent.match(/msie\s([\d\.]+)/i);
+		return match ? parseFloat(match[1], 10) : NaN;
 	}),
 	getHeadElement = memoize(function () {
 		return document.head || document.getElementsByTagName("head")[0];
 	}),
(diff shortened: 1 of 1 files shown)


bencooling:
Was experiencing this exact issue would appreciate a fix!

bebraw:
I would feel a lot more confident in this after merging #176. At least then we should be able to tell that any of the core functionality doesn't inadvertently break as a result.

Another way to approach it would be to design and expose an extension mechanism for **style-loader** where you could plug in support for old IE (less to maintain here), but that might be a little too much.

michael-ciniawsky:
@ bebraw :+1

d3viant0ne:
@jerrysu you will need to get the CLA signed. To do so, you just need to close this PR & open it again. That will trigger the CLA workflow again which should allow you to sign it.

Please note, before you close & open it again ensure the email in your local git config matches what you have configured in github.

jerrysu:
Thanks for accepting! Unfortunately, I don't have the time to update this and verify the fix in IE8 as I have stopped using `style-loader` for my projects. I'll close this, but leave the issue open and let someone that is experiencing this issue to use this fix as a reference.

PR #60, opened 2015-05-06
Wrap sourceMap JSON in unescape and encodeURIComponent.

elektronik2k5:
Fixes error described here: http://stackoverflow.com/a/26603875. I can only reproduce the issue when style-loader is used together with autoprefixer-loader ^1.2.0 in this configuration:
```javascript
config.module.loaders.push({
    test: /\.scss$/,
    loaders: [
      "style-loader",
      "css-loader?sourceMap",
      "autoprefixer-loader",
      "sass-loader?sourceMap&sourceMapContents=true" + helpers.getScssIcludePathsAsArgs(),
    ],
  });
```
Perhaps the issue is with autoprefixer-loader? I'm not sure how to find the root cause.

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -204,7 +204,8 @@ function updateLink(linkElement, obj) {
 	var sourceMap = obj.sourceMap;
 
 	if(sourceMap) {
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #63, opened 2015-05-15
fixes error in ie8

mako-taco:
IE8 Doesn't have Array.prototype.filter; the loader otherwise works amazingly well. If you want to continue functioning in IE8, please consider this PR.

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -157,8 +157,15 @@ var replaceText = (function () {
 	var textStore = [];
 
 	return function (index, replacement) {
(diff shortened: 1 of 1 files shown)


sokra:
It's the job of the application to provide the relevant polyfill. So I won't merge this.

mako-taco:
Unfortunately, if your build target is a library, you cant just polyfill things like this without breaking a million people's crappy `for(x in array)` code that runs in browsers which do not support the enumerable/definiteProperty

PR #65, opened 2015-05-25
Allow to specify element into which to inject Style.

unity:
We use this to allow injecting a Style tag into a different place than HEAD.

We use React+Webpack a lot to bundle reusable applications embedded in a 3rd party page.
Here are two examples of how we benefit from this PR:
- To to inject the style into a virtual Iframe (with https://github.com/ryanseddon/react-frame-component or some other tech we're still working on)
- To inject the style in a specific place in the DOM, so we can monitor it with Mutation Observers


vvo:
HI @unity How do you pass the corresponding element in the webpack config? Thanks

unity:
You pass it at runtime : 

```js
import style from './style.css'

style.use({
  element:document.getElementById('myId');
});
```

mmoutenot:
is there any chance this will be merged? Would love to use this for iframes

ronkorving:
Currently this PR has conflicts. Rebasing would be a good start. Would like to see this merged too.

addStyles.js
```diff
@@ -95,11 +95,11 @@ function listToStyles(list) {
 	return styles;
 }
 
-function createStyleElement() {
+function createStyleElement(element) {
 	var styleElement = document.createElement("style");
```
michael-ciniawsky: `styleElement` => `style`

michael-ciniawsky: `parentElement` => `element`

michael-ciniawsky: and all following references to them please :)

addStyles.js
```diff
@@ -13,9 +13,9 @@ var stylesInDom = {},
 	isOldIE = memoize(function() {
 		return /msie [6-9]\b/.test(window.navigator.userAgent.toLowerCase());
 	}),
-	getHeadElement = memoize(function () {
```
ekulabuhov: I don't see a reason to remove memoization.

addStyles.js
```diff
@@ -13,9 +13,9 @@ var stylesInDom = {},
...
-		return document.head || document.getElementsByTagName("head")[0];
-	}),
+	getElement = function (element) {
```
ekulabuhov: I don't see a reason to remove memoization.

michael-ciniawsky:
Superseded by #135 Thx 😛

Aqours:
```js
import style from './style.css'
style.use(parent.document.getElementById('myId'));
```
IMO, this way is more useful than set config in style-loader. Consider to add this feature

PR #68, opened 2015-06-30
Update README: set publicPath for ?sourceMap to load assets

### LINKED ISSUE #55: Generated image urls *must* be absolute for style!css?sourceMap to work?
Is there a way to make images load in browser when using `style-loader/css-loader?sourceMap` without specifying **absolute** url prefix as `publicPath`? 

With this loader:
```js
loaders.push({
  test: /\.scss$/,
  loader: 'style!css?sourceMap')
});
```
This works:
```js
output = {
  path: 'build',
  publicPath: 'http://0.0.0.0:8080/'
  filename: 'bundle.js'
};
```
This doesn't:
```js
output = {
  path: 'build',
  publicPath: '/'
  filename: 'bundle.js'
};
```
In the CSS modules generated as `Blob` the paths are relative to `chrome:blob` or `chrome:devtools` which doesn't refer to the correct assets. 

If my assumption is correct and we always have to set absolute path in `output.publicPath` I will be happy to add a note about this to the docs.

gryzzly:
fixes #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ So the recommended configuration for webpack is:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

Issue #56, opened 2015-04-24
Allow to specify the position to insert new link/style tag


pkorzh:
i want to be able to specify where to insert new link/style nodes. in my project bower deps. css should go before custom css, witch is built in one bundle and included in head tag.

having query option like `insertAt: 'top' / 'bottom` would do the trick

sokra:
Feel free to send a PR that adds such a query parameter.

Anahkiasen:
:+1: for this

vvo:
This is already done, can be closed

SpaceK33z:
Jup, this was fixed in #69 months ago.

PR #69, opened 2015-07-07
Add support for 'insertAt' parameter

elsbree:
Implemented for #56. Adds an 'insertAt' parameter that lets you specify where you would like to insert the style tags in the head, either 'top' or 'bottom' (the default). This lets you avoid taking priority over any CSS already in the head if you so desire.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ Note: Behavior is undefined when `unuse`/`unref` is called more often than `use`
 
 ### Options
 
+#### `insertAt`
+
+By default, the style-loader appends `<style>` elements to the end of the `<head>` tag of the page. This will cause CSS created by the loader to take priority over CSS already present in the document head. To insert style elements at the beginning of the head, set this query parameter to 'bottom', e.g. `require('../style.css?insertAt=bottom')`.
+
 #### `singleton`
 
 If defined, the style-loader will re-use a single `<style>` element, instead of adding/removing individual elements for each required module. **Note:** this option is on by default in IE9, which has strict limitations on the # of style tags allowed on a page. You can enable or disable it with the singleton query parameter (`?singleton` or `?-singleton`).
--- a/addStyles.js
+++ b/addStyles.js
@@ -29,8 +29,8 @@ module.exports = function(list, options) {
 	// tags it will allow on a page
 	if (typeof options.singleton === "undefined") options.singleton = isOldIE();
 
-    // By default, add <style> tags to the bottom of <head>.
-    if (options.insertAt !== "top" && options.insertAt !== "bottom") options.insertAt = "bottom";
+	// By default, add <style> tags to the bottom of <head>.
+	if (options.insertAt !== "top" && options.insertAt !== "bottom") options.insertAt = "bottom";
 
 	var styles = listToStyles(list);
 	addStylesToDom(styles, options);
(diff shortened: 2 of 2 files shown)


sokra:
please use tabs instead of spaces... :)

elsbree:
Woops, somehow I didn't notice that. I changed the spaces to tabs.

addStyles.js
```diff
@@ -95,11 +98,15 @@ function listToStyles(list) {
...
+function createStyleElement(options) {
 	var styleElement = document.createElement("style");
 	var head = getHeadElement();
 	styleElement.type = "text/css";
-	head.appendChild(styleElement);
+	if (options.insertAt === "top") {
+		head.insertBefore(styleElement, head.firstChild);
+	} else if (options.insertAt === "bottom") {
+		head.appendChild(styleElement);
+	}
```
sokra: `else` throw an error

elsbree: What do you think of this for the error text: "Invalid value for parameter 'insertAt'. Must be 'top' or 'bottom'."

sokra: Sounds good

vvo:
https://github.com/substack/insert-css has "prepend" mode which I do find a good naming. What do you think @elsbree?

elsbree:
Seems like that would mostly be a question of semantics. The 'insertAt' parameter has the advantage of being more flexible, that is, we could allow options other than just 'top' or 'bottom' in the future. That being said, I don't have a strong preference either way.

MattiSG:
+1

README.md
```diff
@@ -45,6 +45,10 @@ Note: Behavior is undefined when `unuse`/`unref` is called more often than `use`
 
 ### Options
 
+#### `insertAt`
+
+By default, the style-loader appends `<style>` elements to the end of the `<head>` tag of the page. This will cause CSS created by the loader to take priority over CSS already present in the document head. To insert style elements at the beginning of the head, set this query parameter to 'bottom', e.g. `require('../style.css?insertAt=bottom')`.
```
sokra:
``` diff
- `require('../style.css?insertAt=bottom')`
+ `require('!style?insertAt=bottom!css!../style.css')`
```

addStyles.js
```diff
@@ -95,11 +98,17 @@ function listToStyles(list) {
...
 	var styleElement = document.createElement("style");
 	var head = getHeadElement();
 	styleElement.type = "text/css";
-	head.appendChild(styleElement);
+	if (options.insertAt === "top") {
+		head.insertBefore(styleElement, head.firstChild);
+	} else if (options.insertAt === "bottom") {
+		head.appendChild(styleElement);
+	} else {
+		throw new Error("Invalid value for parameter 'insertAt'. Must be 'top' or 'bottom'.");
```
sokra: Never reacted because of the check above:

`if (options.insertAt !== "top" && options.insertAt !== "bottom")`

elsbree: @sokra I can remove this line if you think that's cleaner

sokra: better change the check above:

``` diff
- if (options.insertAt !== "top" && options.insertAt !== "bottom") options.insertAt = "bottom";
+ if (typeof options.insertAt === "undefined") options.insertAt = "bottom";
```

an error for incorrect configuration is fine.

elsbree: Just made the change.

sokra:
Thanks

sokra:
Hmm... It's broken, because style tags are in incorrect order.

i. e.

``` js
require("a.css");
require("b.css");
```

bottom:

```html
<head>
  <style src="existing"></style>
  <style>... a.css ...</style>
  <style>... b.css ...</style>
</head>
```

top:

```html
<head>
  <style>... b.css ...</style>
  <style>... a.css ...</style>
  <style src="existing"></style>
</head>
```

and the order is important...

Could you do a follow-up PR that fix this, i. e. by storing a list of all top-inserted style tags and inserting after the last one. Keep in mind that `top` and `bottom` could be mixed.

elsbree:
Oh, good catch. Yep, I can work on that.

PR #76, opened 2015-09-15
feat: add `prepend` query option

vvo:
Much like https://github.com/substack/insert-css prepend option.

Similar to #69. Does not solves #65.

Let me know @sokra


vvo:
thoughts?

sokra:
see #84

vvo:
Nice

PR #77, opened 2015-09-16
Added option to return complete object

rickchristie:
We need this option to write code that can be used in both client and server side.
When building for server side rendering with `css-loader`, locals are accessed this way:

```JavaScript
var locals = require('style.css').locals;
document.write('<div class="' + locals.className + '"></div>');
```

However `style-loader`'s default behavior is to return locals directly, resulting in an error when the above code is run in browser. By using the `returnComplete` option users can opt for a behavior consistent with `css-loader`, making the above code run in both server and client side. 

Please let me know if there are other ways to achieve this.

Thanks,

Rick


sokra:
From the css-loader README.

> Note: For prerendering with extract-text-webpack-plugin you should use css-loader/locals instead of style-loader!css-loader in the prerendering bundle. It doesn't embed CSS but only exports the identifier mappings.

This should work on both, client and server:

``` js
var locals = require('style.css');
document.write('<div class="' + locals.className + '"></div>');
```

server: `css-loader/locals?modules`
client: `style-loader!css-loader?modules`

PR #78, opened 2015-09-16
Add urlPrefix option for absolute URLs.

rickchristie:
##### Absolute URL for assets

Use `urlPrefix` option to transform relative URLs into absolute URLs, in case you wanted to store assets in another domain or use CDN mirror.
For example, if you set `urlPrefix` to `cdn.yourdomain.com`:

```css
.forward {
	background: url('assets/forward-icon.png');
	/* transformed into http(s)://cdn.yourdomain.com/assets/forward-icon.png */
}
```

Please let me know if there's a better way to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,17 @@ require("style/url!file!./file.css");
 // => add a <link rel="stylesheet"> to file.css to document
 ```
 
+### Absolute URL for assets
+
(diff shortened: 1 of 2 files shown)


rickchristie:
Found out about `output.publicPath` configuration!

PR #82, opened 2015-09-27
Exporting locals from css-loader

christianalfoni:
CSS Modules i really awesome (https://github.com/css-modules/webpack-demo), but you can not use it in combination with hot updates. It actually works, but the style-loader does not return the locals from the css-loader.

I have suggested a fix for this. Putting this together with react-transform and you got yourself an insanely powerful combo :-)

Old behaviour:

```js
import styles from './App.css';
styles // {}
```

New behaviour:

Old behaviour:

```js
import styles from './App.css';
styles // {Header: "Header_134135}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ module.exports.pitch = function(remainingRequest) {
 		"	}",
 		"	// When the module is disposed, remove the <style> tags",
 		"	module.hot.dispose(function() { update(); });",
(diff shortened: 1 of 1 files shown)


sokra:
The style-loader already supports CSS Modules with HMR. A line similar to the line you added is already in the file `if(content.locals) module.exports = content.locals;`. Maybe you are using a old version of the style-loader...

Issue #93, opened 2015-11-18
Loading images in css


LestaD:
I try:

``` css
# ./style.css
.searchBg {
  background-image: url(sprite/searchBg.jpg)
}
```

``` js
# ./component.js
import css from './style.css';

// begin component
  <div className={css.searchBg} />
// end component
```

What i see in chrome:
![image](https://cloud.githubusercontent.com/assets/5620073/11242182/32542fdc-8e11-11e5-8b93-3563b09b5abb.png)

But image not loading to background.
When i click on "Open Link in new tab", link opens as:

![image](https://cloud.githubusercontent.com/assets/5620073/11242207/58689ff0-8e11-11e5-9954-77ec78af2c74.png)

Strate:
From README:

> Note about source maps support and assets referenced with url: when style loader is used with ?sourceMap option, the CSS modules will be generated as Blobs, so relative paths don't work (they would be relative to chrome:blob or chrome:devtools). In order for assets to maintain correct paths setting output.publicPath property of webpack configuration must be set, so that absolute paths are generated.

Btw, why style-loader can not replace relative urls to absolute urls by itself?

tizmagik:
See https://github.com/webpack/style-loader/issues/55 for some workarounds/solutions to this issue.

michael-ciniawsky:
Will be fixed by either #96 or #124

PR #96, opened 2015-12-03
Make urls absolute in blob

Strate:
Related #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,8 +70,6 @@ So the recommended configuration for webpack is:
 }
 ```
 
-**Note** about source maps support and assets referenced with `url`: when style loader is used with ?sourceMap option, the CSS modules will be generated as `Blob`s, so relative paths don't work (they would be relative to `chrome:blob` or `chrome:devtools`). In order for assets to maintain correct paths setting `output.publicPath` property of webpack configuration must be set, so that absolute paths are generated.
-
 ## Install
 
 `npm install style-loader`
--- a/addStyles.js
+++ b/addStyles.js
@@ -235,15 +235,15 @@ function updateLink(linkElement, obj) {
 
 	css = css
 		.replace(
-			new RegExp('url(/', 'g'),
+			new RegExp('url\\(/', 'g'),
 			'url(' + absPrefix + '/'
 		)
 		.replace(
-			new RegExp('url(\'/', 'g'),
+			new RegExp('url\\(\'/', 'g'),
 			'url(\'' + absPrefix + '/'
 		)
 		.replace(
-			new RegExp('url(\"/', 'g'),
+			new RegExp('url\\(\"/', 'g'),
 			'url(\"' + absPrefix + '/'
 		)
 	;
--- a/addStyles.js
+++ b/addStyles.js
@@ -235,15 +235,15 @@ function updateLink(linkElement, obj) {
 
 	css = css
 		.replace(
-			'url(/',
+			new RegExp('url(/', 'g'),
 			'url(' + absPrefix + '/'
 		)
 		.replace(
-			'url(\'/',
+			new RegExp('url(\'/', 'g'),
 			'url(\'' + absPrefix + '/'
 		)
 		.replace(
-			'url(\"/',
+			new RegExp('url(\"/', 'g'),
 			'url(\"' + absPrefix + '/'
 		)
 	;
--- a/addStyles.js
+++ b/addStyles.js
@@ -231,6 +231,22 @@ function updateLink(linkElement, obj) {
 	var css = obj.css;
 	var media = obj.media;
 	var sourceMap = obj.sourceMap;
+	var absPrefix = window.location.protocol + '//' + window.location.host;
+
+	css = css
+		.replace(
+			'url(/',
+			'url(' + absPrefix + '/'
+		)
+		.replace(
+			'url(\'/',
+			'url(\'' + absPrefix + '/'
+		)
+		.replace(
+			'url(\"/',
+			'url(\"' + absPrefix + '/'
+		)
+	;
 
 	if(sourceMap) {
 		// http://stackoverflow.com/a/26603875
--- a/addStyles.js
+++ b/addStyles.js
@@ -235,6 +235,18 @@ function updateLink(linkElement, obj) {
 
 	css = css
 		.replace(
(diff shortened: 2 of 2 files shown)


Strate:
Soo?

unimonkiez:
Give this man a cookie!

oller:
Wooo, so does this PR give us full source maps with relative path assets, without the need for an absolute `publicPath`.

:clap:

Strate:
@oller sure. Tested by using in real project, works like a charm.

sergeylukin:
Worked for me at first glance. Any reason why this isn't merged?
@Strate I suggest squashing commits into one

v3lbloud:
Thanks for the patch. Works like charm.

Strate:
@sokra this is really helpful patch, as you can see. could you merge it? Or maybe I could somehow improve this patch?

geekyme:
bump. Need this patch

vperron:
Come on, please merge :)

oller:
Pleeeeaaase! 🎁

tizmagik:
It's been 5 months... Is there any reason for this not to be merged @sokra ?

sokra:
yes. This is not a good solution. It should be solved on CSS AST. It's better to solve it before the style loader in the css loader

codisms:
@sokra I can't find any code in css-loader that is client side, which is where this logic would need to reside so it knows the browser location at run time.  Can you point us in the right direction?

bendytree:
@Strate great idea and great work!

I just submitted a very similar pull request which is opt-in via a `style?fixUrls` flag and has some unit tests to verify the behavior since I'm not parsing the CSS into an AST.

Anyone feel free to fork [bendytree/style-loader](https://github.com/bendytree/style-loader) and then use a git url in your package.json:

    ...
    "style-loader": "git+https://git@github.com/YOUR_USERNAME/style-loader.git",
    ...

tizmagik:
@sokra any guidance you could provide @Strate / @bendytree et al to solve this properly in the CSS AST as you suggested?

Strate:
As for me, use css ast on client side would be an overkill. This patch used in real world project since it's creation without any problem. With keeping in mind that it used only in dev environment, I could say that it has sufficient quality to merge.

SpearThruster:
This is a huge issue on my project as well, can we get update why this hasn't been merged?

motleydev:
Adding an additional, please. Or at least provide feedback how this could be made better.

SimonSelg:
I'll work on this this afternoon. I have some ideas how to solve this correctly on AST level. But I'll have to test the performance impact.

Strate:
**Attention please**.

I've found a solution for the problem without patching this loader.
Just add following line on the top of your topmost entry:
```
__webpack_public_path__ = window.location.protocol + "//" + window.location.host + "/"
```
This line modifies `output.publicPath` option at browser runtime, adding current protocol and host. If you have a custom publicPath, just add it to the end of prefix.
This solution requires you to have no `output.publicPath` specified in config or command line.
If you need to pass publicPath from command line or webpack configuration file you can use environmant variables together with `DefinePlugin`:

```js
// webpack.config.js

module.exports = {
  pluginns: [
    new webpack.DefinePlugin({
      'process.env.WEBPACK_PUBLIC_PATH': JSON.stringify(process.env.WEBPACK_PUBLIC_PATH)
    })
  ]
}

// topmostEntryPoint.js
__webpack_public_path__ = window.location.protocol + "//" + window.location.host + "/" + process.env.WEBPACK_PUBLIC_PATH;
```

and run build with command:
```
WEBPACK_PUBLIC_PATH="your/public/path/" webpack
```

SpearThruster:
@Strate your latest solution doesn't work for me sadly. If I log in runtime the '__webpack_public_path__' I get the correct value but the url properties in css (originally scss) aren't updated. Anyone having this issue?

SpearThruster:
So it's still the css source map issues. The latest workaround by @Strate works at my machine ONLY if I disable source maps. The original pull request works for both scenarios.

ledil:
any news on that ?

Strate:
@SpearThruster, try this:
1. Move you code, which augments `__webpack_public_path__` into separate file, which do only that thing. Call it, for example, `bootstrap-public-path.js`.
2. Add it as your first topmost entrypoint in your webpack configuration:
```
entry: [
  "bootstrap-public-path",
  "index" // your old index entrypoint here
]
```
3. Ensure, that your `webpack.config.js` does not set `publicPath` option.

IAMtheIAM:
@Strate Thank you! This did the trick for me to get the paths to resolve correctly inside my SCSS file. I used it for loading fonts in scss, when the scss file is a "blob://" rather than an inline style. 

```
// topmostEntryPoint.js
__webpack_public_path__ = window.location.protocol + "//" + window.location.host + "/" + process.env.WEBPACK_PUBLIC_PATH;```

d3viant0ne:
Also - There were compliance issues we needed to clear up before the CLA bot could be enabled, that was cleared up about 10 days ago.

To get the CLA signed, you just need to close this PR & open it again. That will trigger the CLA workflow again which should allow you to sign it.

Please note, before you close & open it again ensure the email in your local git config matches what you have configured in github.

ekulabuhov:
I looks like this PR might be superseded by #124?

d3viant0ne:
Yes. My take on all of this would be to land #124, put #165 in the next major & close #96 but that plan needs to be confirmed by the @webpack-contrib/org-maintainers

bebraw:
@d3viant0ne Yeah, sounds fine.

Strate:
@d3viant0ne have you read this comment: https://github.com/webpack-contrib/style-loader/pull/96#issuecomment-228370602? This is how I solve this issue current in my project. I think it is enougth to move this trick to README.

d3viant0ne:
It's workaround vs. fix where the workaround is only effective in some cases. This is an issue that needs to be resolved in code

michael-ciniawsky:
#186 (#124) landed, is this the PR still relevant then? If @Strate please close and reopen to trigger the CLA Bot again and rebase against current master, otherwise if you lost interest in this PR in general / #186 (#124) solved the issue please confirm we can close

d3viant0ne:
screen shot 2017-03-16 at 8 38 29 pm

This should be closed in favor of #124 which has already been merged.

PR #99, opened 2015-12-15
Force repaint in Chrome for link elements

aparticka:


michael-ciniawsky:
@aparticka usecase please and would it only work in Chrome?

michael-ciniawsky:
#121

bebraw:
Please add a comment so people know why the timeout exists.

michael-ciniawsky:
Is the PR already ok the way it is in regard to #121 ? Our is there any detailed work required like add an option to set the delay e.g `options.delay/options.timeout`  in `setTimeOut`?

ekulabuhov:
TBH, I would close this as irrelevant. Too much time passed since the PR was opened. We don't know if the issues is still happening. We don't even know what the issue was to begin with. Doesn't look like anyone else had this issue during that time either.

In relation to #121: I don't see how it could help to solve it as `setTimeout` makes it even more asynchronous then it was.

Are we OK to close this now and re-open if more details emerges in the future?

bebraw:
Ok. Let's close.

@aparticka When you have time, please comment.

PR #112, opened 2016-02-27
Add AMP support

Nivl:
Add support for the [AMP project](https://www.ampproject.org) (which is now used by Google Search).

As stated by the [documentation](https://www.ampproject.org/docs/get_started/create/presentation_layout.html):

> Style elements using class or element selectors in an inline stylesheet in the`<head>`, called `<style amp-custom>` [..] Every AMP page can only have a single embedded stylesheet.

It basically adds a `amp` option (boolean) that activates `singleton` and use `<style amp-custom>` instead of `<style type="text/css">`.


sokra:
I don't think it's valid to inject the style via js in AMP.

michael-ciniawsky:
This should be an generic option to add attrs to the style tag instead. See e.g #119

michael-ciniawsky:
Fixed by #183. Thx nevertheless 😛

PR #115, opened 2016-02-29
fix support server side usage

magalhas:
I'm using local scopes and I wasn't able to render on the server side because of style loader. This little fix solved all my problems :+1:

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -11,7 +11,7 @@ var stylesInDom = {},
 		};
 	},
 	isOldIE = memoize(function() {
-		return /msie [6-9]\b/.test(window.navigator.userAgent.toLowerCase());
+		return typeof window !== 'undefined' && /msie [6-9]\b/.test(window.navigator.userAgent.toLowerCase());
 	}),
 	getHeadElement = memoize(function () {
 		return document.head || document.getElementsByTagName("head")[0];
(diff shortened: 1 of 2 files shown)


randycoulman:
This little fix works great for us as well.  Here's more context on how we're using it: http://randycoulman.com/blog/2016/03/22/testing-with-mocha-and-webpack/ (reference to this issue at the very end, under `Gotchas`).

sokra:
see last section in this chapter: https://github.com/webpack/css-loader#local-scope

use `css/locals` when using SSR.

magalhas:
@sokra does that work if I don't use ExtractTextPlugin and rely on shipping my CSS inside JS? I'm asking because with the PR it worked perfectly, I guess the `css/locals` solution will as well, just want to be sure of it :)

sokra:
`css/locals` just returns the local to class mappings, without style tags, CSS and such stuff, which is not needed for server-side rendering. It works CSS inside JS, but that is usually not used in combination with server-side rendering because you either get FOUC or blocking JS.

magalhas:
I was asking for development. Thank you for clarifying!

sokra:
> I was asking for development.

Then it's fine of course... ;)

peterh32:
So... this never got merged? And style-loader is still broken for server-side rendering?

evilebottnawi:
@peterh32 style-loader works good with SSR. Please don't write about your problem in old closed issue,
create a new issue with description and minimally reproducible test repo. Thanks!

alsun:
It look like in 1.0.0 it's come back :cry:

evilebottnawi:
@alsun no

PR #122, opened 2016-05-11
Fixed negative refs

tomasAlabes:
Style should be able to be injected if `unuse` was called more times than `use`.

```
myCss.use()
myCss.unuse()
myCss.unuse()

myCss.use() //should be valid, regardless of the double call to unuse
```

Following the GC idea, there are not "negative" amount of references, just none. Preventing adding more refs makes no sense.
This basically prevents `refs` from being negative, then the behavior is the same as before.


tomasAlabes:
@sokra any feedback on this?
Thanks

bebraw:
I would say #176 needs to go in first. After that adding a test for this PR would be possible. 👍

bebraw:
Alright. #176 is in. Please add a test and we'll merge.

tomasAlabes:
@bebraw test is in place. I ran it without my change and it failed, with my change 🚀 
Thanks for your work in the loaders guys!

PR #132, opened 2016-07-08
Break the `sourceMappingURL` string to avoid confusing Chrome.

kring:
We recently started seeing "Failed to load SourceMap" problems loading in our app in Chrome 51.  I eventually tracked it down to this line in `style-loader`'s `addStyles.js`:

```
css += "\n/*# sourceMappingURL=data:application/json;base64," + btoa(unescape(encodeURIComponent(JSON.stringify(sourceMap)))) + " */";
```

I suspect that Chrome is incorrectly interpreting this as an actual sourcemap reference.  In any case, breaking up the `sourceMappingURL` solved the problem for us, so I hope you might consider merging this simple workaround!

For what it's worth, the problem does not occur in Chrome Canary, but that's not expected to hit stable until September or October.

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -232,7 +232,9 @@ function updateLink(linkElement, obj) {
 
 	if(sourceMap) {
 		// http://stackoverflow.com/a/26603875
-		css += "\n/*# sourceMappingURL=data:application/json;base64," + btoa(unescape(encodeURIComponent(JSON.stringify(sourceMap)))) + " */";
+		// The seemingly-unnecessary break between 'source' and 'Mapping' is to prevent Chrome 51 (at least) from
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Related #181

michael-ciniawsky:
@kring Current direction seems to be solving this in `source-map-loader` itself, please join discussion in #181

ekulabuhov:
Filed against Chrome 51. It seems to be fixed now since version 52 (according to https://github.com/davezuko/react-redux-starter-kit/issues/858#issuecomment-234822927). Probably can be closed?

michael-ciniawsky:
Closed in favour of #181 or direct fix in `source-map-loader` also proposed in #181

PR #133, opened 2016-07-08
fix: avoid adding duplicate styles

Kovensky:
In a case where:
- Entry point includes A
- CSS file A includes B and C
  - CSS file C includes D
    - CSS file D also includes B

`addStylesToDom` would see B as having two identical parts, and thus it would get added twice to the head.

Removing the `else` branch here stops it from happening, and works locally.

I am not sure what the `parts` handling is for, though. In my case, the `list` has two instances of the same dependency (`list[0] === list[1]`), which results in duplicate inclusion; so it just results in adding two of the exact same style tag to the DOM.

I may also be doing something dumb, though for now I am merely testing a complex config rather than writing real css with that hierarchy. The `parts` handling was added at the same time as deduplication was, but I suppose the deduplication only works with linear dependency graphs.

In what case would `parts` with `length >= 1`  have something that is _not_ a duplicate?

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -93,8 +93,6 @@ function listToStyles(list) {
 		var part = {css: css, media: media, sourceMap: sourceMap};
 		if(!newStyles[id])
 			styles.push(newStyles[id] = {id: id, parts: [part]});
-		else
-			newStyles[id].parts.push(part);
 	}
 	return styles;
 }


michael-ciniawsky:
👋  😛 Please close and reopen to trigger the CLA bot again + a quick test would be nice :D

michael-ciniawsky:
@Kovensky What's the status here ? 😛

bebraw:
@Kovensky Did you think about a test?

michael-ciniawsky:
@Kovensky last time I will bug you with this 🙃 , do we need a test here and do you have something in mind? Otherwise I need to close this  😛

Kovensky:
@michael-ciniawsky the thing is, I don't remember anymore the conditions I got this to happen, or how I could reproduce it in a test.

I _think_ it might have been related to one file that only had `:root { --custom-var }`  declarations, which get stripped by `postcss-css-variables` on defaults.

michael-ciniawsky:
So this maybe was more of a project specific issue then a common bug in `style-loader` ? 😛  If so we close this for now and in case it happens again simply make a new PR then, could you nevertheless please post a small example of the file (setup) and what wasn't deduped ? 🙏  I will make a note about it just in case 🙃

michael-ciniawsky:
Feel free to reopen when we know how to test this

Issue #44, opened 2015-02-27
Passing document to .use


nwhite89:
I need to be able to pass the document to the `.use` method in which it should attach the document, due to iFrame buster business. Any way of doing this currently or thought on perhaps adding as an added feature?

Thanks.

solsson:
+1
```javascript
var style = require("style/useable!css!./embed.css");
$('#myIframe')[0].onload = function() {
  // TODO add style to this frame, which is same origin
};
```

michael-ciniawsky:
Related to #135

michael-ciniawsky:
Closed by #135

PR #126, opened 2016-05-24
[Enhancement] Add parentElement parameter to use/ref method

Craga89:
Based on https://github.com/webpack/style-loader/pull/65, rebased onto current master and cleaned up some-what. This allows the user to specify an optional `parentElement` to inject the `link`/`style` tag into at runtime, using the `use`/`ref` method:

``` javascript
var style = require("style/useable!css!./file.css");
var iframe = document.getElementsByTagName('iframe')[0];
style.use(iframe.contentDocument.head); // = style.ref(iframe.contentDocument.head);
style.unuse(); // = style.unref();
```

Useful for those of us rendering into `iframe`s!

PR #135, opened 2016-07-13
Add insertInto option

### LINKED ISSUE #131: insertAt shadow root?
I'm curious if this community has given any thought to use cases involving shadow elements.

What I'm envisioning is a capability that would allow you to inject styles into a shadow root, as opposed to just the head of the document, thereby preserving encapsulation for a given component.

I am admittedly new to web components, but using css loader to inject encapsulated styles into shadow elements just sounds like an awesome capability :) 

Anyone given thought to this use case?

Thanks!

ahstro:
Add another option called `insertInto` - which allows specifying a CSS selector - e.g. `#styles` - to be used as an argument to `document.querySelector` to find an element into which to insert the generated `<style>`-tags. Will default to the `<head>`-tag if undefined.

This was created because I wanted to insert `<style>` tags into a ShadowRoot (like in #131), but this is not a specific solution for that, but does solve it by allowing people to first create a ShadowRoot and then target it using `insertInto` and the appropriate selector.

Not att all sure if this is wanted, but it exists, so there ya go :)
Also, I replaced the `getHeadElement` function with a custom memoized function rather than add the key feature to the existing memoizing function, not sure if that was the better choice, but it felt like it.

Closes #131

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,10 @@ Note: Behavior is undefined when `unuse`/`unref` is called more often than `use`
 
 #### `insertAt`
 
-By default, the style-loader appends `<style>` elements to the end of the `<head>` tag of the page. This will cause CSS created by the loader to take priority over CSS already present in the document head. To insert style elements at the beginning of the head, set this query parameter to 'top', e.g. `require('../style.css?insertAt=top')`.
+By default, the style-loader appends `<style>` elements to the end of the style target, which is the `<head>` tag of the page unless specified by `insertInto`. This will cause CSS created by the loader to take priority over CSS already present in the target. To insert style elements at the beginning of the target, set this query parameter to 'top', e.g. `require('../style.css?insertAt=top')`.
+
+#### `insertInto`
+By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else, e.g. into a [ShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot), you can specify a CSS selector for that element here, e.g. `require('../style.css?insertInto=#host::shadow>#root')`.
 
 #### `singleton`
 
--- a/addStyles.js
+++ b/addStyles.js
@@ -13,8 +13,16 @@ var stylesInDom = {},
 	isOldIE = memoize(function() {
 		return /msie [6-9]\b/.test(self.navigator.userAgent.toLowerCase());
 	}),
-	getHeadElement = memoize(function () {
-		return document.head || document.getElementsByTagName("head")[0];
+	getElement = (function(fn) {
+		var memo = {};
+		return function(selector) {
+			if (typeof memo[selector] === "undefined") {
+				memo[selector] = fn.call(this, selector);
+			}
+			return memo[selector]
+		};
+	})(function (styleTarget) {
+		return document.querySelector(styleTarget)
 	}),
 	singletonElement = null,
 	singletonCounter = 0,
@@ -33,7 +41,10 @@ module.exports = function(list, options) {
 	// tags it will allow on a page
 	if (typeof options.singleton === "undefined") options.singleton = isOldIE();
 
-	// By default, add <style> tags to the bottom of <head>.
+	// By default, add <style> tags to the <head> element
+	if (typeof options.insertInto === "undefined") options.insertInto = "head";
+
+	// By default, add <style> tags to the bottom of the target
 	if (typeof options.insertAt === "undefined") options.insertAt = "bottom";
 
 	var styles = listToStyles(list);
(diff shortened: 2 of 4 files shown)


geekuillaume:
Is there any news on this PR ?
@ahstro you need to accept the CLA to merge this.

Thanks for your work !

ahstro:
@geekuillaume The "Details" link dosn't work, says "There is no CLA to sign for webpack/style-loader". Same is true when I change `webpack` in the URL to `webpack-contrib`.

rustydb:
Can anyone sign this?

ahstro:
Signing the CLA worked now, so if there's nothing I need to change, this should be good to merge now.

danielverejan:
So is there anyone reviewing PRs?

addStyles.js
```diff
@@ -13,8 +13,16 @@ var stylesInDom = {},
 	isOldIE = memoize(function() {
 		return /msie [6-9]\b/.test(window.navigator.userAgent.toLowerCase());
 	}),
-	getHeadElement = memoize(function () {
-		return document.head || document.getElementsByTagName("head")[0];
+	getStyleTarget = (function(fn) {
```
michael-ciniawsky: `getStyleTarget` => `getElement/getStyle`

ekulabuhov:
This is very similar to https://github.com/webpack-contrib/style-loader/pull/65.

ahstro:
@bebraw I'm looking at testing, but the existing test utils seem to be built around this feature not existing, by comparing only the contents of the `<head>` tag with the expected result. The solution I came up with is comparing the entire jsdom document source with a string gotten from a new method on `utils` that inserts the expected style output at some index in the `jsdomHtml`, but that might be a really bad way of doing things and I'm tired, so I'm not going to work more on that now. Hope you or someone approves or, preferably, has a better idea soon. :)

bebraw:
@ekulabuhov Any idea on what would be a good way to test the change?

ekulabuhov:
@ahstro I think you might need to replace that document.head reference with the querySelector. Add path as the last parameter to runCompilerTest and default it to head. 

Come to think about it, I'm not sure if querySelector can query Shadow DOM.

Ping me if you have any troubles adding this.

ahstro:
@ekulabuhov We're using my fork for work to query a ShadowDOM, so it works, you just need some fancy query.

I don't have access to the computer where I started work on testing yesterday, so I can't fix it right now, but I'll try to get to it later today.

michael-ciniawsky:
For `v1.0.0` I'm currently drafting a few refactors for dicsussion (will open a detailed issue later :D) and one of the things I would lije to change is e.g

Merge `insertAt/insertTo/...`  => `selector: 'head' || 'body' || '#el' || '.el'` (use `querySelector` + fallback) like in this PR

README.md
```diff
@@ -47,9 +47,12 @@ Note: Behavior is undefined when `unuse`/`unref` is called more often than `use`
 
 ### Options
 
+#### `insertInto`
```
michael-ciniawsky: Please put it directly below the `insertAt` section

addStyles.js
```diff
@@ -13,8 +13,16 @@ var stylesInDom = {},
...
+	getElement = (function(fn) {
+		var memo = {};
+		return function(selector) {
+			if (typeof memo[selector] === "undefined") {
+				memo[selector] = fn.call(this, selector);
+			}
+			return memo[selector]
+		};
+	})(function (styleTarget) {
+		return document.querySelector(styleTarget)
```
michael-ciniawsky: We need a fallback here, I guess 😛  ?

**Pseudo Fallback**
```js
if (!document.querySelector) {
  if (selector.match('#'))  return document.getElementById(selector)
  if (selector.match('.'))  return document.getElementByClassName(selector)
  return document.getElementByTagName(selector)
}  else {
  return document.querySelector(selector)
}
```

michael-ciniawsky:
Also please take a look into https://github.com/webpack-contrib/style-loader/pull/65/files (4) adding it to useable aswell 🙃

ekulabuhov:
@michael-ciniawsky did you mean #126 in your last comment? 😀

I like the idea of merging `insertAt` and `insertTo` 👍

addStyles.js
ahstro: Do we really? According to Can I Use, the support for [querySelector](http://caniuse.com/#search=querySelector) in newer browsers is better than that of [getElementsByClassName](caniuse.com/#search=getElementsByClassName), the latter not having any support for IE8 while the former has partial support.

README.md
ahstro: On it :)

ahstro:
I do like the idea of merging them, but if I understood it correctly, that's out of scope for this PR, right?

michael-ciniawsky:
@ahstro Yep sry for making noise about it here, this will be `v1.0.0`  => `InsertTo` for next minor

ahstro:
@michael-ciniawsky No problem :) Did you see the comment I left on your line comment regarding the fallback?

addStyles.js
michael-ciniawsky: Forget what I said then

michael-ciniawsky:
Closes #44

michael-ciniawsky:
Closes #131

ahstro:
Okay, added tests :) I kept it a separate commit so I can easily change anything if it's not satisfactory; let me know if I should squash them.

ahstro:
Changed default parameter syntax because Travis complained

jinyang1994:
How to use this feature? I want insert into multiple iframe.

PR #137, opened 2016-07-25
Server side ability

janoist1:
A split point has been added that gives ability to this loader to work on the server side. On client, everything is the same but on server side it won't try to manipulate the DOM and also, it has been extended to collect the styles into a global variable `global.__styles__` which can be used when assembling the layout.


ndreckshage:
@janoist1 using a singleton for ssr? this seems to build styles between page requests.

`/page-1` -> `.page1 { color: blue}`
`/page-2` -> `.page1 { color: blue} .page2 { color: green}`
`/page-3` -> `.page1 { color: blue} .page2 { color: green} .page3 { color: purple}`

ekulabuhov:
Both this and https://github.com/webpack-contrib/style-loader/pull/159 are trying to accomplish similar goals. From what I gather, SSR is outside the scope for the `style-loader`. I think we need to add that to the Readme file and direct people to common solutions for this problem.

PR #144, opened 2016-10-01
Replace window with self for better web worker support

oskarer:
Replacing `window` with `self` to better support web workers, they are the same thing in UI thread; https://developer.mozilla.org/en-US/docs/Web/API/Window/self.

This enables style loading in web worker context, for example by using this technique; https://github.com/web-perf/react-worker-dom/tree/r-15.

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -11,7 +11,7 @@ var stylesInDom = {},
 		};
 	},
 	isOldIE = memoize(function() {
-		return /msie [6-9]\b/.test(window.navigator.userAgent.toLowerCase());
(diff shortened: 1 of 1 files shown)


TheLarkInn:
This makes sense I wish we had a test suite to back this up.

SpaceK33z:
@TheLarkInn, replacing `window` with `self` does no harm. All browsers support it, and it's the same as `window`, with the difference that it's read-only. Recently we did the same for webpack-dev-server, and nobody complained.

Thanks for your PR @oskarer.

Issue #130, opened 2016-06-09
Conditional Stylesheet


braden-qualtrics:
It would be nice if there was an option to insert conditional stylesheets. So put them on the page like this:

```
<!--[if IE]>
    <link rel="stylesheet" type="text/css" href="all-ie-only.css" />
<![endif]-->
```

Maybe something like this would work:

```
require("style?condition=ie!./mycss.css");
```

michael-ciniawsky:
Related #146

michael-ciniawsky:
Discussion moved to #146

Issue #37, opened 2015-01-14
Do something after styles have been updated


gaearon:
I want to apply viewport units polyfill on iOS 7 as soon as styles are added dynamically.

It's important I want to do this at runtime because it's optional — and I use `ExtractTextPlugin` so can't post-process all CSS on client side.

What would be the easiest way to do something after styles have been flushed?
I don't *really* care about reflows here because it's only for iOS 7.

gaearon:
FWIW this works well for me:

```js
'use strict';

var viewportUnitsBuggyfill = require('viewport-units-buggyfill'),
    { any } = require('underscore');

viewportUnitsBuggyfill.init();
var patchedViewportStyleNode = document.getElementById('patched-viewport');

function isAppStyleNode(node) {
  return node.tagName === 'STYLE' && node !== patchedViewportStyleNode;
}

function refreshViewportBuggyfill() {
  viewportUnitsBuggyfill.refresh();
  patchedViewportStyleNode.parentNode.appendChild(patchedViewportStyleNode);
}

function handleHeadMutation(mutationRecords) {
  if (any(mutationRecords, record => any(record.addedNodes, isAppStyleNode))) {
    refreshViewportBuggyfill();
  }
}

if (patchedViewportStyleNode) {
  if (typeof MutationObserver !== 'undefined') {
    new MutationObserver(handleHeadMutation).observe(document.head, {
      childList: true
    });
  } else {
    setInterval(refreshViewportBuggyfill, 1500);
  }
}
```

This is probably rather unusual request so closing this.

seanemmer:
Thanks for posting this.  I'm attempting to run Viewport Units Buggyfill as part of a similar Webpack build and am unable to successfully apply the viewport units to my targeted DOM elements.

Had a couple of follow up questions:

1) What is the 'patched viewport' referenced here?  Do I need to re-run this script for each DOM element I am applying CSS viewport properties to (in certain views there are multiple nested elements)?

2) My app is structured as an SPA.  Do I need to wait until the target DOM elements are rendered to execute the script above, or can I execute the script once when the application initializes.

gaearon:
Sorry, I don't remember anything about this 😔

PR #146, opened 2016-10-27
feat: add option.transform

brafdlog:
I added the option to pass the name of a function that can make changes to the css before adding it to the html. The function can return false, in which case the css won't be loaded at all.
The relevant use cases for this are changes that that need context that is not available at build time.

For example:
1. (partial) support for css variables in IE (which is the use case I currently need)
2. Conditional stylesheet loading (issue #130)
3. A similar function can be used for issue #37 (perhaps a postCssAdded callback)

These things were very hard (impossible?) to do without having a hook that gives control of the css when it is loaded into the browser.

My initial thought was to add another loader but since this transformation needs to run at runtime I need to take css as input and output javascript. So I can't put it before the style loader (which expects css, not js) and cant put it after the style loader (because then I would have to somehow parse it's javascript and change it which is quite brittle and not very elegant)

There is one thing that I don't like about this solution and that is that you need the `transformCssOnLoad` function to be available on the global scope for this to work. I could not think of a better way to support this, if there is a better idea I would love to hear it.


brafdlog:
@sokra What do you think?

addStyles.js
```diff
@@ -143,6 +143,20 @@ function createLinkElement(options) {
 function addStyle(obj, options) {
 	var styleElement, update, remove;
 
+	// If a transformCssOnLoad function was defined, run it on the css
+	if (options.transformCssOnLoad && obj.css) {
```
michael-ciniawsky: `options.transformCssOnLoad` =>`options.transform`

addStyles.js
```diff
@@ -143,6 +143,20 @@ function createLinkElement(options) {
...
+		var transformationFunction = self[options.transformCssOnLoad];
```
michael-ciniawsky: `var tranformationFunction` => `const/let transformation`

ekulabuhov:
Isn't that what loaders are for - transformation? Why put a loader inside a loader?

addStyles.js
```diff
@@ -143,6 +143,20 @@ function createLinkElement(options) {
...
+		if (!transformationFunction) {
+			throw new Error("A function named " + options.transformCssOnLoad + " needs to be available on the global scope.");
```
ekulabuhov: A function on global scope seems dirty. Is it possible to pass a function instead of a name here?

bebraw:
@ekulabuhov If I'm interpreting right, there's a coupling problem since `addStyles` is something loaded indirectly. That said, if it's possible to push the idea to a separate loader outside of this project, that would be better.

michael-ciniawsky:
> The relevant use cases for this are changes that that need context that is not available at build time.

@ekulabuhov Let's wait for a response, I'm also not 💯 yet

addStyles.js
brafdlog: I chose this name to make it clear that this transformation occurs at load time. I think this name makes it very clear what this function does. Do you think transform is better?

brafdlog:
@ekulabuhov @d3viant0ne 

> My initial thought was to add another loader but since this transformation needs to run at runtime I need to take css as input and output javascript. So I can't put it before the style loader (which expects css, not js) and cant put it after the style loader (because then I would have to somehow parse it's javascript and change it which is quite brittle and not very elegant)

Am I missing something here? How would I do this with another loader?

d3viant0ne:
@brafdlog - You aren't missing anything with style-loader in it's current state. At the moment, there is no good way to extract this but at the same time, it really should be.

My point, before we just hammer this home I would prefer to try and find a way to not shoehorn a loader into a loader. I get what you are trying to do and why, we just need to find a better way to accomplish it.

ekulabuhov:
@brafdlog - @d3viant0ne explained it pretty well. Just to clarify, I do think it's a great feature to have. But I, as a webpack user, wouldn't expect to find it in a style-loader. 

I wonder, could you solve these issues using refs use() / unuse() ? Or would it be too cumbersome?

michael-ciniawsky:
Ok naysayers how would a solution with a different loader look like ? 😛

bebraw:
The tricky bit is that **addStyles.js** is a script loaded by **style-loader**. Maybe the question is actually would it be possible to add some kind of an extension point there? This might help with other functionality too.

ekulabuhov:
@michael-ciniawsky: `npm i magical-runtime-loader -D` :)

But seriously, is it possible to have 2 pitch loaders in one chain?

michael-ciniawsky:
Add the option to inject a `addStylesRuntime.js/addStylesContext.js` file, but I'm fine with adding a `transform/runtime` function to `addStyles.js`

michael-ciniawsky:
> But seriously, is it possible to have 2 pitch loaders in one chain?

@ekulabuhov ? 🙃

ekulabuhov:
@michael-ciniawsky https://webpack.github.io/docs/loaders.html#pitching-loader
If I understand correctly, `style-loader` works by injecting code rather than data. Is it possible to wrap `style-loader` code within another `runtime-transform-loader`?

Also found a way to pass an object instead of query string to a loader: https://webpack.github.io/docs/how-to-write-a-loader.html#programmable-objects-as-query-option

liady:
@michael-ciniawsky @brafdlog We can achieve the separation of concerns by doing the following:
1. Build a new dedicated 'runtime-css-transform' loader - that receives raw css as input and returns a js function (stringified) that, when invoked, returns transformed css.
2. Extend the style-loader to receive a css string *or* a stringified js function (that returns css). If it receives a function - it runs it (on **runtime**, not build time) and operates as usual on the returned css.

Then a user can apply them in this order:
raw css => css-loader => runtime-css-transform-loader (that returns a function)=> style-loader (that runs this function)

What do you think?

p.s
We can make the 'runtime-css-transform' loader generic, to operate on any arbitrary string, not just css.

michael-ciniawsky:
fixes #146

addStyles.js
bebraw: `transformCssOnLoad` sounds fine to me. `transform` alone does not communicate the intent.

michael-ciniawsky:
@brafdlog @bebraw @d3viant0ne @ekulabuhov @liady I still think it isn't a bad idea to support passing a `transform/runtime` function to `style-loader` directly, we need to find a final agreement on this. Please vote with reactions on this comment ( :+1: / :-1: )

michael-ciniawsky:
@brafdlog In any case please sign the CLA by closing and reopening the PR to trigger the CLA Bot again :)

addStyles.js
michael-ciniawsky: kk. `runtime` ? `transformOnLoad`?

bebraw: `transformOnLoad` sounds like a good compromise. Not too long but it also communicates something.

brafdlog:
@michael-ciniawsky @bebraw @d3viant0ne @ekulabuhov @liady
I think this pull request adds new and useful abilities and solves a few pending issues. 
There are a number of comments here about things in the solution that are not optimal. 
If there are concrete ideas that we agree on that can improve the solution, lets do it. Otherwise I think it would be a pity not to merge this.
Once we agree on the solution I will add the documentation, tests etc.

addStyles.js
brafdlog: `transformOnLoad` and `transformCssOnLoad` both sound good.

bebraw:
About CLA bot, it checks Git commits against GitHub email so make sure both match.

brafdlog:
@bebraw I learned that the hard way :)
Fixed it so now the checks pass.

addStyles.js
michael-ciniawsky: @brafdlog Yep the sole `transform` is to jaded, personally in favour of `transformOnLoad`, the more precise and short the better 😛 . I will quit bikesheding about naming from this point, choose based on majority and preference :)

d3viant0ne:
`Loaders should do a single job. The only job of the style-loader is to add style to the DOM - Tobais Koppers`

The question isn't if there is a problem that needs solving, it's should that problem be solved in style-loader.  The answer to that question is no, it should not. `This simply does not belong here`.

If there is a community need that goes beyond workarounds for the failings of Internet Explorer, we can spin up a defaults based project within `webpack-contrib` to address it.

liady:
@d3viant0ne What do you think about the solution I proposed earlier (for separation of concerns)?
https://github.com/webpack-contrib/style-loader/pull/146#issuecomment-285315451

michael-ciniawsky:
As it would be a opt-in, I think it's bikesheding on when a function is actually a 'loader' or not, the same could be said for `css-loader`'s  `cssBase.js` implementation which splits ths css into dedupable chunks => `css-base-loader` ? Also in terms of UX I personally don't think it is more straigthforward or intentional/intuitive to apply another loader for this (extra boilerplate, extra loader invocation, not compatible with current `style-loader` && `extract-text-webpack-plugin`*)

Maybe Tobias should chime in here and give is general blessing 😛 or reject it, because of a better alternative approach

* for the latter it doesn't have to but I'm afraid folks will use it ignorantly

michael-ciniawsky:
> I would favor a separate loader, but I guess the option is ok too, as it doesn't require much runtime code. But don't use a global var. Maybe a module name as option which is required by the loader is a better way. Or abstract it to select a custom style injector, defaulting to the normal one (alternatives: singleton injector, insertInto injector, etc.)

@brafdlog We can move on if we can do it like Tobias described in the comment above 👍

d3viant0ne:
I'm still in the separate loader camp but I've said my piece and the great OZ has spoken. 

So if we do this as a part of style-loader, there needs to be a clearly communicated, defined set of use cases that will be supported. This particular feature would be super easy to start abusing.

We all want to avoid this becoming bloated or a perf issue 6 months down the road.

brafdlog:
@michael-ciniawsky @d3viant0ne I made a few changes as discussed (and some additional that were not discussed yet :) )

1. Renamed transformationFunction -> onCssLoad - the logic for this is that this can actually be used as a general hook, not necessarily a transformation. If you return a new value, this will be a transformation but if the function returns nothing, it will load the original css. If the function returns false the css will not be loaded - conditional loading.
2. Instead of using a global var, pass a module name to the loader (as @sokra said).
3. Documentation
4. Tests

README.md
```diff
@@ -79,6 +79,31 @@ If defined, the style-loader will re-use a single `<style>` element, instead of
 
 If convertToAbsoluteUrls and sourceMaps are both enabled, relative urls will be converted to absolute urls right before the css is injected into the page. This resolves [an issue](https://github.com/webpack/style-loader/pull/96) where relative resources fail to load when source maps are enabled.  You can enable it with the convertToAbsoluteUrls query parameter (`?convertToAbsoluteUrls`).
 
+#### `onCssLoad`
+
+If an `onCssLoad` function is defined, this function will run every time a css is loaded into the page through the style-loader.
+The `onCssLoad` function gets one parameter that is the original css that the style-loder is loading into the page.
+Possible return values of `onCssLoad`:
+* **nothing** - if nothing (undefined) is returned from the `onCssLoad` function, the original css will be loaded without any changes.
```
michael-ciniawsky: nothing => `undefined`

README.md
```diff
@@ -79,6 +79,31 @@ If defined, the style-loader will re-use a single `<style>` element, instead of
...
+The `onCssLoad` function gets one parameter that is the original css that the style-loder is loading into the page.
+Possible return values of `onCssLoad`:
+* **nothing** - if nothing (undefined) is returned from the `onCssLoad` function, the original css will be loaded without any changes.
+* **false** - if the `onCssLoad` function returns **false**, the css will *not* be loaded into the page. 
+* **updated css** - If the function returns a value, this value will be added into the page instead of the original css. This can be used to run transformations on the loaded css.
+
+Usage:
+
+```javascript
+// myCode.js - define an onCssLoad function
```
michael-ciniawsky: Move the comment above the code block please  e.g

 **transformCSS.js**
```js
module.exports = function (css) {
   const transformed = // Here we can change the original css;
   // e.g const transformed = css.replace('.classNameA',  '.classNameB')
   return transformed; // If false is returned the css will not be loaded.
}
```

README.md
```diff
@@ -79,6 +79,31 @@ If defined, the style-loader will re-use a single `<style>` element, instead of
...
+module.exports = function(originalCss) {
+    var transformed = // Here we can change the original css;
+    return transformed; // If false is returned the css will not be loaded.
+}
+```
+
+```javascript
```
michael-ciniawsky: \```javascript  => \```js

README.md
michael-ciniawsky: Maybe a table is better here :)

|Name|Type|Default|Description|
|:----:|:----:|:------:|:-----------|
|**`undefined`**|`{}` (?)|`true`|If nothing (undefined) is returned from the `onCssLoad` function, the original css will be loaded without any changes|
|**`false`**|`{Boolean}`|`false`|If the `onCssLoad` function returns **false**, the css will *not* be loaded into the page|
|**`transformed`**|`{Function}`|`undefined`|If the function returns a value, this value will be added into the page instead of the original css. This can be used to run transformations on the loaded css|

It's a bit confusing where && what from a first glance tbh, these are all the possible the return values of `onCssLoad`, but especially the first one `undefined` seems to be unnecessay. If you don't set `options.onCssLoad` this is the default anyways 😛

README.md
michael-ciniawsky: Also please add an example for conditional loading after `tranformCSS.js`

 **conditionalCSS.js**
```js
module.exports = function (css) {
   // If the condition is matched load the CSS
   if (css.includes('/* onCssLoad */')) {
     return css; 
   }
   // By returning `false` the CSS won't be loaded 
   return false
}
```

README.md
```diff
@@ -79,6 +79,31 @@ If defined, the style-loader will re-use a single `<style>` element, instead of
...
+require('style-loader?onCssLoad=path/to/onLoadFunction/myCode.js!style.css');
```
michael-ciniawsky: **webpack.config.js**
```js
{
  loader: 'style-loader'
  options: {
    onCssLoad: (css) => {} || require('path/to/transformCSS')
  }
}
```

d3viant0ne:
@michael-ciniawsky - In this particular case, after the usual suspects are all satisfied I'd like Tobias to take a look at this one as well before it's merged.

README.md
brafdlog: @michael-ciniawsky The `nothing`/`undefined` option is a convenience. The logic behind it is that if the user wants to write some code that runs on css load but doesn't change the css (a hook, not a transformation) I won't force him to return the original css. Since this seems complicated (more than it actually is) I will remove this option to keep it simple.

michael-ciniawsky: @brafdlog Please provide a example for dicussion, but it seems this is out of scope and would need a particular usecase :)

michael-ciniawsky:
@brafdlog What's the status here ? 😛

brafdlog:
@michael-ciniawsky I just recentley had a baby so I am working on his timetable :)
I will push the changes as soon as I can.

michael-ciniawsky:
@brafdlog ohh.. 🙃  congrats, well take your time 😛

brafdlog:
@michael-ciniawsky I pushed fixes for the comments you had.
I simplified the flow by removing the option to return nothing. So now the rule is, if the value you return from the function is truthy, it will be loaded instead of the original css. Otherwise, the css will not be loaded at all.
Since this is the case, calling the function onCssLoad is now misleading so I renamed it back.

brafdlog:
@bebraw @ekulabuhov @d3viant0ne Are there any other changes we need to make?

README.md
```diff
@@ -79,6 +79,47 @@ If defined, the style-loader will re-use a single `<style>` element, instead of
...
+    cssTransformation: 'path/to/transformCSS'
+  }
+}
+```
+
+**transformCSS.js**
+```js
+module.exports = function(originalCss) {
+    // Here we can change the original css
+    const transformed = css.replace('.classNameA',  '.classNameB');
```
bebraw: `originalCss` over `css` so that the code works.

bebraw: Or you can fix the other way around. Then it's consistent with the code below.

michael-ciniawsky: `css` > `originalCss`  please 😛

michael-ciniawsky:
Please rename to `options.transform` I requested the same in #164 it's simply terser and the `css` prefix doesn't add any special meaning (this loader only 'processes' CSS anyways) 

```
transform: (css) => {...}
```
should be self explanatory and the option is well documented, finally good to go after that 😛

PR #153, opened 2016-12-01
refactor: support force disabling sourcemaps and setting sourceRoot

specialkk:

feature

**Summary**

SourceMaps have so far been a real pain to track down the source of issues with, and the way CSS loaders handle them seem to be all over the place. Most of these stem from the `sourceRoot` property being set and then integrated into the actual source paths somewhere down the line. This is especially problematic if webpack isn't being run in the current working directory (as I am). As long as it's blank throughout though, things seem to work out.

However, once the maps arrive at `style-loader`, they have no root, and without one they get put under the `(no domain)` section in Chrome's Sources view. The `ExtractTextPlugin` handles these fine since the paths are set via `moduleFilenameTemplate`. I could write an intermediate loader just to set `sourceRoot` to `webpack:///`, but it would be nice if I could do that through `style-loader`'s config, which is the goal of this PR.

Nope!

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -171,12 +171,17 @@ function attachTagAttrs(element, attrs) {
 function addStyle(obj, options) {
 	var styleElement, update, remove;
 
+	if (obj.sourceMap && typeof options.sourceRoot === "string") {
+		obj.sourceMap.sourceRoot = options.sourceRoot;
+	}
+
 	if (options.singleton) {
 		var styleIndex = singletonCounter++;
 		styleElement = singletonElement || (singletonElement = createStyleElement(options));
 		update = applyToSingletonTag.bind(null, styleElement, styleIndex, false);
 		remove = applyToSingletonTag.bind(null, styleElement, styleIndex, true);
 	} else if(obj.sourceMap &&
+		options.sourceMap !== false &&
 		typeof URL === "function" &&
 		typeof URL.createObjectURL === "function" &&
 		typeof URL.revokeObjectURL === "function" &&


michael-ciniawsky:
@rockmacaca  Can you please Close and reopen the PR  to trigger the CLA Bot and rebase against current master ? 😛

d3viant0ne:
Before anyone considers / begins to review this pull request, it needs to be rebased with current master and it has to be `rebase`

rockmacaca:
I'll rebase when I have a chance later this week!

rockmacaca:
@michael-ciniawsky this is ready to be looked at! Sorry I didn't mention it earlier

addStyles.js
```diff
@@ -171,12 +171,17 @@ function attachTagAttrs(element, attrs) {
 function addStyle(obj, options) {
 	var styleElement, update, remove;
 
+	if (obj.sourceMap && typeof options.sourceRoot === "string") {
```
michael-ciniawsky: @rockmacaca This is not mandatory, more of a triage 😛 , but would it be a breaking change, if we change the signature of `sourceMap`  to `sourceMap: Boolean || { sourceRoot: '...' }` instead of adding a new option `sourceRoot` ? Shouldn't be, but needs to be verified

bebraw: Yeah, I would rather go with this than to add yet another field to the API. Better reuse this time.

michael-ciniawsky:
@rockmacaca According to Tobias `sourceRoot` should always be empty to work with webpack, could you elaborate again on why you want to modify it ? 😛

michael-ciniawsky:
@rockmacaca friendly ping 😛

rockmacaca:
Sorry, I've been putting this off for a bit too long -

Yes, `sourceRoot` is expected to be empty as maps get passed through webpack. The last time I checked, I believe css-loader typically used in conjunction will take care of setting the `webpack:///` prefix on maps, but I am choosing to use my own set of loaders. If you're curious what it is, it's just

sass-loader --> postcss-loader --> raw-loader (modified with sourcemap support) --> style-loader  

For production builds, I have style-loader swapped with `ExtractTextPlugin` (which doesn't have the problem this PR addresses since it uses the `moduleFilenameTemplate` function). As the paths shouldn't be prefixed with anything by the time they arrive at style-loader, they appear in the `(no domain)` section. I'm making the PR here since it shouldn't be the responsibility of any intermediate loader to correctly set the prefix.

michael-ciniawsky:
Agreeing upon that and `postcss-loader` v2.0.0 (especially for Angular folks) will export a module directly (no `raw-loader`) in the near future, could you post an usage example of how you would to use this feature purposed in the PR and could we fix it without introducing an extra option for it (based on a pattern etc)? 😛

rockmacaca:
Sure. This is a rule I have for my plain CSS files:
```javascript
{ use: ['style-loader?sourceRoot=webpack:///', 'css-raw-loader'], include: path.resolve(__dirname, './app') }
```
css-raw-loader is literally just a copy of raw-loader but it attaches a sourcemap.

I like being able to just use the inline query string format, but your suggestion of having a signature of
`sourceMap: Boolean || { sourceRoot: '...' }` would also work. To be honest, the original `sourceMap` option I added should probably be removed entirely. I only added that while I was still figuring out how webpack's loader lifecycle worked :P. Since the entire sourcemap is still included in the bundle and all it is accomplishing is hiding it from the browser's sources inspector, it isn't really useful.

addStyles.js
michael-ciniawsky:
```js
if (obj.sourceMap && !obj.sourceMap.sourceRoot.match(/webpack:\/\//) {
  obj.sourceMap.sourceRoot = 'webpack://' + obj.sourceMap.sourceRoot
}
```

michael-ciniawsky: @rockmacaca Something in that direction is what I mean by 'can we avoid adding an extra option'? 🙃 If not possible and the user needs control over this, then it can't be helped 😛

michael-ciniawsky:
@rockmacaca How should we proceed with this ?

rockmacaca:
I'll let you know in a bit, I think I'm just going to remove the boolean `sourceMap`. This PR is pretty tiny so I can rebase on top of #219 if that's getting merged soon.

michael-ciniawsky:
Yep no rush, so when #219 lands, this will still be a needed? Tbh I haven't had much time recently to take a deeper look at it, so sry in advance if this is bit naively asked 😛

rockmacaca:
Ok, I just realized what that PR does and completely fixes the reason I added the `sourceMap` option since I use relative paths in my CSS. I still want to be able to customize `sourceRoot` though. I'll amend this sometime soon and I'll just resolve any conflict that arises once 219 is taken care of.

michael-ciniawsky:
Sry but I have to decline this, since source maps will be inlined in `style-loader >= v1.0.0`, this PR isn't needed/won't work in `v1.0.0`.

specialkk:
That's all right, I haven't been too concerned about it. It's actually pretty simple to create my own loader that handles prefixing the paths.

In the meantime, are there plans to publish a beta or RC for v1.0 to npm anytime soon?

michael-ciniawsky:
Yes `v1.0.0-alpha` will be released in ~1-2 days, maybe today if nothing else pops up on my schedule :)

PR #169, opened 2017-02-04
Remove license link

simon04:
To allow inclusion in https://webpack.js.org/ docs.

See https://github.com/webpack/webpack.js.org/pull/804


bebraw:
Note that if the license links are added back, they should be absolute (fixes both webpack site and npm).

PR #173, opened 2017-02-22
Add href replace option.

danielverejan:
Feature: replace href attribute of the <link> that points to the required stylesheet.

No tests

Updated the README to include the replace option but it may need additional input.

**Summary**
In short: it works similar as the url option, the difference is that it look for the `<link>` that has a reference to the required stylesheet instead of creating one.

I'm working on a .Net project where the views are Razor template, so no options there to use InterpolateHtmlPlugin because the views are static. Nor I can use the style-loader with or without the url option for dynamic style loading(HMR) as the view references are static to. By static references I mean that ex: the `<link rel="stylesheet" href="Build/css/index.css">` is used in DEV and PROD environments.

No breaking changes.


michael-ciniawsky:
@danielverejan I'm in favour of the feature, but not of the implementation tbh :). But also no idea how to do it atm

@bebraw @d3viant0ne @ekulabuhov  Any ideas ? 😛

d3viant0ne:
Code diving aside, there is a valid need though I question the scope of this particular features usefulness within the scope of style-loader. It's a pretty niche requirement

bebraw:
I agree with @d3viant0ne. This might work better as a loader of its own. Given how the code has been written, that would not be an issue.

ekulabuhov:
Similar opinion. The code is very specific to this particular problem. @danielverejan would it work for you if you moved it into a separate loader and then chain it as usual?

michael-ciniawsky:
Yep a seperate package e.g `link-loader` is an option

bebraw:
Do you think `link-loader` could live below contrib? I can set up a repo based on **webpack-defaults** and invite Daniel there.

d3viant0ne:
That is the path I would suggest.

bebraw:
Only problem - [link-loader](https://www.npmjs.com/package/link-loader) is taken.

@danielverejan Suggest a name available at npm and I'll set you up.

Time to close this PR. 👍

michael-ciniawsky:
Accepting under webpack-contrib would be even better, very nice 😛 

ahhh...npm naming, maybe even ask the owner to choose another name for his package :D

danielverejan:
Thanks for the many responses.
@michael-ciniawsky  I have at work a version with some miner improvements like `querySelector` and would be happy to get any suggestion for further  improvements.

@d3viant0ne  Working on a .Net project, before webpack I have used browser-sync which has basically this feature, it's called there something like *style injection*. Being forced to use a plain `<style>` tag for dev and prod this was very useful. That's what I have been missing when I moved to webpack with style-loader.

I'll come in with a name shortly, thanks for the opportunity!

danielverejan:
@bebraw Maybe some of these?

restyle-loader
relink-loader
css-link-loader

bebraw:
@danielverejan Sure. Pick the one you like the most and I'll set you up. 👍

danielverejan:
restyle-loader

bebraw:
@danielverejan Done. I set up that follows **webpack-defaults** setup. You'll have to fill in meta and publish. Add me and @d3viant0ne as npm project owners (`npm owner add`) after you have published something. 👍

danielverejan:
ok, thanks!

PR #175, opened 2017-02-24
Adds type attribute to the generated link element

dimitarivanov:
Although the rel element is set to stylesheet the created link element does not specify the MIME type of the linked resource.
The MIME type is set to text/css similar to the case when an style element is created. This change also syncs with the addStylesUrl.js behavior where the type attribute is set.

bugfix/enhancement

no

no

**Summary**
There are tools/frameworks that rely on MIME type for processing static content prior delivering the resources. In this particular case the href tag is not updated correctly because the type attribute is missing.


michael-ciniawsky:
@dimitarivanov  Thx 👍

PR #176, opened 2017-02-27
Added unit tests for basic configurations

ekulabuhov:
**What kind of change does this PR introduce?**
Build related change

**Did you add tests for your changes?**
Yes

**If relevant, did you update the README?**
Not relevant

**Summary**
This PR introduces unit testing setup we can use to check for regressions when adding new features to style-loader. Hopefully, this can provide a base for PR authors to test against. 

The setup consists of: 
* [memory-fs](https://github.com/webpack/memory-fs) to avoid writing to file system and to collocate input files
* [jsdom](https://github.com/tmpvar/jsdom) to provide lightweight DOM implementation

Each test is running fully setup webpack compiler. This allows for great level of flexibility. Included are 5 tests for basic configurations shown in the Readme file.

**Does this PR introduce a breaking change?**
No

**Other information**

PR #181, opened 2017-03-03
Avoid source-map-loader matches incorrect source map url

zuohaocheng:

bugfix

No. There is no test in the repository.

Not relevant.

**Summary**

There is a library bundled by webpack with style-loader. A consumer tries to load it through webpack with source-map-loader, but it fails to build.

The root cause is source-map-loader incorrectly thinks the string literal `"\n/*# sourceMappingURL=data:application/json;base64,"` is a directive for source map, it tries to find the source map here(while the actual source map directive is in the end of file), and fails.

This PR provide a workaround on the issue.

No.

**Other information**

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -245,7 +245,7 @@ function updateLink(linkElement, obj) {
 
 	if(sourceMap) {
 		// http://stackoverflow.com/a/26603875
-		css += "\n/*# sourceMappingURL=data:application/json;base64," + btoa(unescape(encodeURIComponent(JSON.stringify(sourceMap)))) + " */";
+		css += "\n/*# sourceMapping" + "URL=data:application/json;base64," + btoa(unescape(encodeURIComponent(JSON.stringify(sourceMap)))) + " */";
 	}
 
 	var blob = new Blob([css], { type: "text/css" });


zuohaocheng:
It seems there is something wrong in the CI config, which thinks the project is in Ruby.

May anyone kindly take a look?

ekulabuhov:
IMO, this should be fixed in source-map-loader instead.

bebraw:
> IMO, this should be fixed in source-map-loader instead.

Yeah, that could be a better direction. 👍

michael-ciniawsky:
Related #132

michael-ciniawsky:
@bebraw @ekulabuhov kk 👍

zuohaocheng:
@ekulabuhov @bebraw @michael-ciniawsky Thanks for review. Actually I'm working on fix in source-map-loader, and it should be done soon.

Meanwhile, it would still be nice to have a workaround here, to keep it from confusing other non-standard compliant consumers, like mentioned in #132

michael-ciniawsky:
@zuohaocheng  How far from the finishing line is your `source-map-loader` fix? I promise to keep track of it and update here approbately (asap), if your/a fix is provided in `source-map-loader`, but would like to avoid any temporarily changes if possible 🙃

michael-ciniawsky:
Also please rebase, so added test in #176 can run at least

zuohaocheng:
@michael-ciniawsky I've rebased and CI passed, thanks for reminding. 

I'm expecting to finish work on `source-map-loader` in a day or two, and will let you know when PR created there.

d3viant0ne:
Given this has been fixed in webpack-contrib/source-map-loader#31 this should / can be closed now yes?

michael-ciniawsky:
I'm not 💯, but hopefully yes :D @zuohaocheng Can you confirm? I will also take a look into it later

zuohaocheng:
@d3viant0ne @michael-ciniawsky Yes it is fixed in `source-map-loader`. I think it is ok for my case - though I'm still a bit worrying about other non-standard consumers.

d3viant0ne:
@bebraw @ekulabuhov - Personal opinion here is to contend with edge cases as the arise and not add a workaround for a non-existant problem which may have unintended side effects.

bebraw:
Yeah, maybe close this one given `source-map-loader` was fixed. If edge cases come up, fix then without trying to pre-empt problems that might not exist.

I think the bigger problem is that there's no standard test suite for testing anything source map related. Though tackling that might be more up to standards bodies and browser authors than webpack. 👍

PR #155, opened 2016-12-10
Add support for custom tag attribute

felixmosh:
**What kind of change does this PR introduce?**
feature, allows to add additional attributes on the style / link element that the loader creates

**Did you add tests for your changes?**
No, tests
**If relevant, did you update the README?**
Update README
**Summary**

This PR allows to add id attribute or any custom attribute, in my case i have runtime css style processor that relies on custom attribute on the style tag that it should process. 

**Does this PR introduce a breaking change?**
no breaking changes

**Other information**

Fixes #147

PR #183, opened 2017-03-06
feat(tag-attribute): Add support for custom tag attribute

### LINKED ISSUE #147: load css with media=all attribute
Hey there,

I initially posted the following issue to: https://github.com/webpack/css-loader/issues/294
Someone told me I would be better served posting it here instead. Do you think you can help me? 
Here is the issue:

I'd like to add the media=all attribute to the css files that webpack is generating:
image

The reason for this is because the css is not being used when I try to print my webpack-generated webpage. 

Thanks for the help!

felixmosh:
What kind of change does this PR introduce?
feature, allows to add additional attributes on the style / link element that the loader creates

Did you add tests for your changes?
No, tests
If relevant, did you update the README?
Update README
Summary

This PR allows to add id attribute or any custom attribute, in my case i have runtime css style processor that relies on custom attribute on the style tag that it should process.

Does this PR introduce a breaking change?
no breaking changes

Other information

Fixes #147
Related #155


d3viant0ne:
ship_it

PR #124, opened 2016-05-20
Add Option To Fix Relative Urls

bendytree:
When sourceMaps are enabled, relative urls stop working. This appears to be a pain point for a lot of people. ([Ref 1](https://github.com/webpack/style-loader/pull/96), [2](https://github.com/webpack/style-loader/issues/93), [3](https://github.com/webpack/style-loader/issues/55), [4](https://github.com/webpack/css-loader/issues/29), [5](https://github.com/webpack/file-loader/issues/29), [6](https://github.com/webpack/webpack-dev-server/issues/226), [7](https://github.com/coryhouse/react-slingshot/issues/81), [8](https://github.com/webpack/css-loader/issues/216), [9](http://stackoverflow.com/questions/31549268/webpack-css-loader-not-finding-images-within-url-reference-in-an-external-styl?rq=1), etc)

Basically `style-loader` uses a link data-uri to embed the css with a source map. That means relative urls have no way to resolve their path. 

Ideally urls should be converted to absolute urls before the `style-loader` runs, but that isn't always possible.  In my opinion `style-loader` is in a unique position to alleviate this issue.

This pull request creates an opt-in option `fixUrls`. When enabled, it uses regex to replace relative image/font urls with absolute urls according to the page location.

Ideally I'd parse the css, but client-side parsers look to be 40-80KB and would likely be a lot slower. Parsing has to happen client side because the location is not known until then.

I've included about 30 test (which can be run with `npm test`) to verify that relative urls are resolved properly, absolute urls are not touched, etc.

PR #165, opened 2017-01-16
refactor: modifying source map support to also support relative urls in stylesheets

john-bai:
Fixing broken relative and protocol-relative url() use in stylesheets when source maps are enabled. Chrome Canary 57.0.2981.0 and Firefox 50.1.0 support loading source maps when sourceMapURL directive is dynamically added to a new or existing style element in the DOM. This means that style-loader can simply add the styles and sourceMapURL directive as text inside a style element and the source map will be loaded by the browser. Given the simpler insertion method, the relative and protocol-relative urls in stylesheets will not break.

Plenty of discussion on this bug in #124. This PR would be great to merge in once browser support is at an acceptable level. Currently Chrome Canary 57.0.2981.0 and Firefox 50.1.0 work with this source map support.

README updated to remove the warning about source maps and relative urls in styles breaking.

PR #186, opened 2017-03-09
added option to fix relative urls

sontek:
This is rebased from bendytree (#124)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,10 @@ By default, the style-loader appends `<style>` elements to the end of the `<head
 
 If defined, the style-loader will re-use a single `<style>` element, instead of adding/removing individual elements for each required module. **Note:** this option is on by default in IE9, which has strict limitations on the number of style tags allowed on a page. You can enable or disable it with the singleton query parameter (`?singleton` or `?-singleton`).
 
+#### `fixUrls`
+
+If fixUrls and sourceMaps are both enabled, relative urls will be converted to absolute urls right before the css is injected into the page. This resolves [an issue](https://github.com/webpack/style-loader/pull/96) where relative resources fail to load when source maps are enabled.  You can enable it with the fixUrls query parameter (`?fixUrls`).
+
 #### `attrs`
 
 If defined, style-loader will attach given attributes with their values on `<style>` / `<link>` element.
@@ -113,7 +117,7 @@ So the recommended configuration for webpack is:
 }
 ```
 
-**Note** about source maps support and assets referenced with `url`: when style loader is used with ?sourceMap option, the CSS modules will be generated as `Blob`s, so relative paths don't work (they would be relative to `chrome:blob` or `chrome:devtools`). In order for assets to maintain correct paths setting `output.publicPath` property of webpack configuration must be set, so that absolute paths are generated.
+**Note** about source maps support and assets referenced with `url`: when style loader is used with ?sourceMap option, the CSS modules will be generated as `Blob`s, so relative paths don't work (they would be relative to `chrome:blob` or `chrome:devtools`). In order for assets to maintain correct paths setting `output.publicPath` property of webpack configuration must be set, so that absolute paths are generated. Alternatively you can enable the `fixUrls` option mentioned above.
 
 <h2 align="center">Contributing</h2>
 
--- a/README.md
+++ b/README.md
@@ -72,9 +72,9 @@ By default, the style-loader appends `<style>` elements to the end of the `<head
 
 If defined, the style-loader will re-use a single `<style>` element, instead of adding/removing individual elements for each required module. **Note:** this option is on by default in IE9, which has strict limitations on the number of style tags allowed on a page. You can enable or disable it with the singleton query parameter (`?singleton` or `?-singleton`).
 
-#### `fixUrls`
+#### `convertToAbsoluteUrls`
 
-If fixUrls and sourceMaps are both enabled, relative urls will be converted to absolute urls right before the css is injected into the page. This resolves [an issue](https://github.com/webpack/style-loader/pull/96) where relative resources fail to load when source maps are enabled.  You can enable it with the fixUrls query parameter (`?fixUrls`).
+If convertToAbsoluteUrls and sourceMaps are both enabled, relative urls will be converted to absolute urls right before the css is injected into the page. This resolves [an issue](https://github.com/webpack/style-loader/pull/96) where relative resources fail to load when source maps are enabled.  You can enable it with the convertToAbsoluteUrls query parameter (`?convertToAbsoluteUrls`).
 
 #### `attrs`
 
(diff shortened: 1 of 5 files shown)


d3viant0ne:
@sontek - You will need to sign the CLA.

@bebraw @michael-ciniawsky - We are going with this in favor of #124 as a prelude to @john-bai's work in #165 which is waiting on chrome-canary updates to propagate down.

fixUrls.js
```diff
@@ -0,0 +1,58 @@
...
+
+	//convert each url(...)
+	var fixedCss = css.replace(/url *\( *(.+?) *\)/g, function(fullMatch, origUrl){
+		//strip quotes (if they exist)
+		var unquotedOrigUrl = origUrl
+			.replace(/^"(.*)"$/, function(o,$1){ return $1; })
+			.replace(/^'(.*)'$/, function(o,$1){ return $1; });
+
+		//already a full url? no change
+		if (/^(data:|http:\/\/|https:\/\/|file:\/\/\/|\/\/)/i.test(unquotedOrigUrl))
```
d3viant0ne: Please check the comment from @john-bai regarding urls with hash fragments https://github.com/webpack-contrib/style-loader/pull/124#issuecomment-272723968

@romulof fixed this error here: https://github.com/romulof/style-loader/commit/c695210c5acbc9a9bbcf826829744ccf056922d3

sontek:
@d3viant0ne Yeah, I signed it but I'm not sure if @bendytree tree has.  How do you check that?

bebraw:
Good point. @bendytree probably has to sign too unless his commit is lost from the history (probably something to avoid until we hear back from him).

sontek:
Only thing I don't like about the PR from @romulof is that it moves from taking a `currentUrl` arg to using `window` directly.   What is your opinion on that? Do you like the change?

bebraw:
The `window` bit will fair server side most likely. But do people use **style-loader** on server side? It doesn't make a lot of sense there.

sontek:
Ok, I pulled in the code and updated the tests to mock `window.location` to work with it but it looks like the code is incomplete.  It didn't provide a test to prove it was working but it seems to have broke the existing tests as well. 

If someone wants to play with getting the @romulof code working, the branch of it is here:

https://github.com/sontek/style-loader/pull/1

But its doing things like `location.protocol + location.host` which won't work because `location.protocol` returns without `//` in it.

addStyles.js
```diff
@@ -35,6 +36,9 @@ module.exports = function(list, options) {
 	// By default, add <style> tags to the bottom of <head>.
 	if (typeof options.insertAt === "undefined") options.insertAt = "bottom";
 
+	// Fix urls defaults to false
+	if (typeof options.fixUrls === "undefined") options.fixUrls = false;
```
ekulabuhov: There was a discussion in a previous PR about this. It seems like it makes sense to enable this by default. Otherwise we get broken CSS and no advice from Webpack on how to fix it.

Just to be safe and not break previous workarounds we could add few checks: 
1. that sourceMaps are enabled 
2. `output.publicPath` is not set

ekulabuhov:
@sontek thanks for the PR. Could you also copy the README changes and the unit tests.

I can see a commit that says "Mochaify unit tests" but it looks you've just deleted the contents? 🤔

sontek:
How should hash URLs work?

if you get `#bg.jpg` do you fix it or just leave it as `#bg.jpg`?

sontek:
How should double slash URLs work?  if you have `//a/b.jpg` do you add protocol onto it?

sontek:
@ekulabuhov Tests are there now, I haven't moved README in yet because there is discussion of defaulting it to `true` which means the docs won't be correct.

sontek:
For my questions above:

both `//` urls and `#` urls stay untouched.

sontek:
@d3viant0ne I believe I've made all the changes you requested.

@ekulabuhov I've made most of the changes you've requested.  Only one I'm not sure of is how to check if `publicPath` is defined.  I have it auto turning `fixUrls` on if they haven't set `fixUrls` and they have `sourceMap` enabled though.

fixUrls.js
sontek: Fixed!  I added a test to show this change,  It doesn't touch any hash URLs now

d3viant0ne:
Definitely want @bebraw to take a look at this one as well before it goes anywhere.

d3viant0ne:
For reference, @sontek is the only one that needs to sign the CLA. This is a new branch with a single committer, the dead pull request from the user in question will simply be closed.

The comment about the hash fix is related to @john-bai's comment in the dead pull request who has this problem well researched. 

As far as using `window`, this will fail server side which may cause issues in `angular-universal` or any other server side rendering.

README.md
```diff
@@ -72,6 +72,10 @@ By default, the style-loader appends `<style>` elements to the end of the `<head
 
 If defined, the style-loader will re-use a single `<style>` element, instead of adding/removing individual elements for each required module. **Note:** this option is on by default in IE9, which has strict limitations on the number of style tags allowed on a page. You can enable or disable it with the singleton query parameter (`?singleton` or `?-singleton`).
 
+#### `fixUrls`
+
+If fixUrls and sourceMaps are both enabled, relative urls will be converted to absolute urls right before the css is injected into the page. This resolves [an issue](https://github.com/webpack/style-loader/pull/96) where relative resources fail to load when source maps are enabled.  You can enable it with the fixUrls query parameter (`?fixUrls`).
```
bebraw: Based on the description it feels like there could be a better name for the parameter. How about `convertToAbsoluteUrls`? `fixUrls` feels too vague and you have to look it up to understand what it fixes exactly.

ekulabuhov:
@d3viant0ne on the `window` debate: `style-loader` is a runtime loader. For SSR the best way to go - is to capture all CSS statically at build-time. `extract-text-plugin` seems to be the most common way to do it. Currently `style-loader` will intentionally throw an exception if used outside the browser.

sontek:
@michael-ciniawsky I believe I've made all the changes you requested as well

README.md
sontek: Is general consensus that `convertToAbsoluteUrls` is the way we are going?   I have no strong opinions about this as long as I get the feature so I can stop using a fork :)

ekulabuhov:
I would prefer #165 because it removes code, decreasing the complexity and chance for bugs creeping in. But let's check if #165 actually works before closing this one.

sontek:
Yeah, my only concern is that this PR comes with a pretty large amount of tests that prove what its doing is correct and will work.  It also works with new any version of Chrome.  #165 hasn't been proven to work, comes with no tests, and only works with 57+

d3viant0ne:
Yeah, that's pretty much what I am thinking @sontek. Realistically, #165 is going to take some time to update and properly vet. 

This issue has been a pain point for a non-trivial amount of time and this pull request solves a problem in the current `0.x.x` version range and allows the chrome version to lag a little bit without requiring everyone to update to what should probably a `Major` based on the browser requirement as Evergreen doesn't necessarily mean up to date.

We can push in this PR & any other of value to this Major range, then start merging in features / fixes that will be included in the `1.0.0`. Currently that  includes ...

1.) #165 will need to be updated to account for what is going to change here & then merged.
2.) Upgrade to `webpack-defaults` 
3.) Any viable PR currently open & of value.

`1.0.0-beta.0` will then begin release validation and be published on the `@beta dist-tag`.

PR #192, opened 2017-03-15
Match parens recursively on URLs to not fix embeded calls

### LINKED ISSUE #191: Minification issue

When build in minification, it appears that in certain scenarios an additional `"` is added at the end of an `rgba` declaration. It is very hard to reproduce with a simple example, but confirmed that the built version in the file has a standard declaration (such as `rgba(0,0,0,0.5)`) and when injected into the DOM it adds a `"` at the end, e.g. `rgba(0,0,0,0.5")`. It is sparingly, so it is tough to pinpoint an exact cause, but I have confirmed with eyeball reconciliation that between the stored version in the file and the rendered version in the DOM this injection occurs. Sorry I can't offer more details, but reverting to 0.13.2 it operates correctly, so I wonder if the "relative URL" fix that was added somehow created this.

**If the current behavior is a bug, please provide the steps to reproduce.**

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**

webpack version = 2.2.1
node version = 6.9.4
os version = manjaro linux

sontek:
Fixes #191 / adds test

### DIFF
--- a/fixUrls.js
+++ b/fixUrls.js
@@ -29,10 +29,35 @@ module.exports = function (css) {
   var currentDir = baseUrl + location.pathname.replace(/\/[^\/]*$/, "/");
 
 	// convert each url(...)
-	var fixedCss = css.replace(/url\(((?:[^)(]+|\((?:[^)(]+|\([^)(]*\))*\))*)\)/g, function(fullMatch, origUrl) {
-		// strip quotes (if they exist)
+	/*
+	This regular expression is just a way to recursively match brackets within
+	a string.
 
+	 /url\s*\(  = Match on the word "url" with any whitespace after it and then a parens
+	   (  = Start a capturing group
+	     (?:  = Start a non-capturing group
+	         [^)(]  = Match anything that isn't a parentheses
+	         |  = OR
+	         \(  = Match a start parentheses
+	             (?:  = Start another non-capturing groups
+	                 [^)(]+  = Match anything that isn't a parentheses
+	                 |  = OR
+	                 \(  = Match a start parentheses
+	                     [^)(]*  = Match anything that isn't a parentheses
+	                 \)  = Match a end parentheses
+	             )  = End Group
+              *\) = Match anything and then a close parens
+          )  = Close non-capturing group
+          *  = Match anything
+       )  = Close capturing group
+	 \)  = Match a close parens
+
+	 /gi  = Get all matches, not the first.  Be case insensitive.
+	 */
+	var fixedCss = css.replace(/url\s*\(((?:[^)(]|\((?:[^)(]+|\([^)(]*\))*\))*)\)/gi, function(fullMatch, origUrl) {
+		// strip quotes (if they exist)
 		var unquotedOrigUrl = origUrl
+			.trim()
 			.replace(/^"(.*)"$/, function(o, $1){ return $1; })
 			.replace(/^'(.*)'$/, function(o, $1){ return $1; });
 
--- a/fixUrls.js
+++ b/fixUrls.js
@@ -29,8 +29,9 @@ module.exports = function (css) {
   var currentDir = baseUrl + location.pathname.replace(/\/[^\/]*$/, "/");
 
 	// convert each url(...)
-	var fixedCss = css.replace(/url *\( *(.+?) *\)/g, function(fullMatch, origUrl) {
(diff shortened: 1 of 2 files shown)


sontek:
@ekulabuhov @michael-ciniawsky @tquetano-r7 Please review

michael-ciniawsky:
@bebraw @ekulabuhov ping 😛

fixUrls.js
```diff
@@ -29,8 +29,9 @@ module.exports = function (css) {
   var currentDir = baseUrl + location.pathname.replace(/\/[^\/]*$/, "/");
 
 	// convert each url(...)
-	var fixedCss = css.replace(/url *\( *(.+?) *\)/g, function(fullMatch, origUrl) {
+	var fixedCss = css.replace(/url\(((?:[^)(]+|\((?:[^)(]+|\([^)(]*\))*\))*)\)/g, function(fullMatch, origUrl) {
```
d3viant0ne: pro-sauce regex :)

bebraw: Yeah. It would be better for the next coder if you split that up a bit (even in a comment). Scary to read.

sontek: Would it be acceptable to bring in a dependency instead? This regex is only necessary because standard regex in the browser doesn't support recursive matching.

https://www.npmjs.com/package/xregexp  could do this in a cleaner manner.  

I'll try to explain it in a comment but its basically just a nasty way to recursively match brackets

bebraw: It would be fine to me (simpler code is better, also less to maintain this way). Better wait what others think.

d3viant0ne: That lib itself has no dependencies, is small and looks to be well maintained. Normally I'd say just use the regex but he has a point about browser regex being a bit gimp.

bebraw: Ok, let's go with a dep then. 👍

sontek: I added a comment on there that might clear it up so we don't need the dep, what do you think?

fixUrls.js
```diff
@@ -29,9 +29,35 @@ module.exports = function (css) {
...
+	             )  = End Group
+              *\) = Match anything and then a close parens
+          )  = Close non-capturing group
+          *  = Match anything
+       )  = Close capturing group
+	 \)  = Match a close parens
+
+	 /gi  = Get all matches, not the first.  Be case insensitive.
+	 */
+	var fixedCss = css.replace(/url\s*\(((?:[^)(]|\((?:[^)(]+|\([^)(]*\))*\))*)\)/gi, function(fullMatch, origUrl) {
```
ekulabuhov: Well, this got complicated pretty fast 😄 

Another options to visualise it: https://regexper.com/#%2Furl%5Cs*%5C(((%3F%3A%5B%5E)(%5D%7C%5C((%3F%3A%5B%5E)(%5D%2B%7C%5C(%5B%5E)(%5D*%5C))*%5C))*)%5C)%2Fgi

sontek: That is a pretty awesome site!

ekulabuhov:
Great work 👍  I think it's good to go.

fixUrls.js
ekulabuhov: @sontek If you think that xregexp could make this simpler, then go with it. The breakdown you did is useful, but the intent is kinda of lost there.

ekulabuhov:
Noticed previous comments:
@sontek If you think that xregexp could make this simpler, then go with it. The breakdown you did is useful, but the intent is kinda of lost there.

sontek:
#193 is to visualize how it'd look with `xregexp`

sontek:
The code is simpler but in general I don't mind this PR, and it always sucks to add an extra dep

michael-ciniawsky:
I would also favour to not include an extra dependency, a comment should be enough, maybe include a link to a regexp visualisation service and good to go imho. (But haven't a strong opinion upon that) #192 #193 are both fine

sontek:
I vote #192 to avoid the dep, we have this ticket + the comment in the PR to document the regex.  Once you squint at it a bit its not that bad :)

fixUrls.js
d3viant0ne: It's how I survive dealing with any regex beyond the basics, second bookmark on my chrome bar.

michael-ciniawsky: 🤣 http://regexr.com / http://regexr.com/3fhgn here. Glad not to be the only one

PR #193, opened 2017-03-16
Match parens recursively on URLs to not fix embeded calls

sontek:
Alternative option to #192 using xregexp


michael-ciniawsky:
@sontek Thy again 😛 Appreciated

Issue #177, opened 2017-02-27
ReferenceError: self is not defined ( Mocha+JSdom )


dlin-me:
## Version 

0.13.2

## Request

Bug

## Current behavior

Got the following error when running test cases with mocha+jsdom. **Same test run fine with version 0.13.1**

```
ReferenceError: self is not defined
    at eval (Users/dlin/Projects/insiight/onboarding/node_modules/style-loader/addStyles.js:14:30)
    at eval (Users/dlin/Projects/insiight/onboarding/node_modules/style-loader/addStyles.js:9:47)
    at module.exports (Users/dlin/Projects/insiight/onboarding/node_modules/style-loader/addStyles.js:31:68)
```

ekulabuhov:
Hi! Do you mind providing a bare bone repro of your issue?

frodare:
This patch broke our testing build.  After looking into the problem we were able to fix it by defining a global self variable in our jsdom setup like this:

```
var jsdom = require('jsdom');
global.document = jsdom.jsdom('<!DOCTYPE html><head></head><html><body></body></html>');
global.window = document.defaultView;
global.window.document = global.document;
// fix
global.self = global.window;
```

ekulabuhov:
This issue was caused by https://github.com/webpack-contrib/style-loader/commit/9dc45a6dee16553a64071031d18feb1562df86a2

The problem here is that `self` is only defined in Browsers and Web Workers. Unit tests running in NodeJs environment won't have access to `self`.

FlyingDR:
Same issue breaks attempts to use `style-loader` to build Firefox add-ons (at least SDK-based) since `self` here contains no `navigator` property and it results into `TypeError: self.navigator is undefined` error

ekulabuhov:
`style-loader` is designed to be used in the browser only. Both `self` and `navigator` are standard properties of HTML5 spec. While it is possible to make `style-loader` work in other environments, it is not something that is supported or guaranteed to work.

FlyingDR:
@ekulabuhov Since the only barrier for use of `style-loader` into other environments is a test for old IE browsers - what stops this test from returning `false` for these non-standard environments which are not old IE by itself?

Updating [this](https://github.com/webpack-contrib/style-loader/blob/master/addStyles.js#L14) line of code with something like
```return window && document && document.all && !window.atob;```

[...]

PR #196, opened 2017-03-21
Update for test for old IE versions

FlyingDR:
Bugfix for #177

Change is only testable in a browsers, code itself was tested using Browserstack into whole range of IE versions across all versions of Windows starting from XP and into outdated / actual versions of all browsers on Windows / OS X / Android / iOS / Windows Phone.

Doesn't seems to be relevant

**Summary**
Test for old IE versions is updated to avoid blind accessing properties of global objects that may not be available into certain real environments (examples are [here](https://github.com/webpack-contrib/style-loader/issues/177#issue-210376738) and [here](https://github.com/webpack-contrib/style-loader/issues/177#issuecomment-284250644)). Since the whole purpose of this test is about giving positive result for old IE versions (<=IE9) it can be safely updated with a [test](http://browserhacks.com/#hack-e71d8692f65334173fee715c222cb805) that is specifically targeted to these versions and doesn't tries to get deeper into (potentially missing) properties of globals.

No

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -16,7 +16,7 @@ var stylesInDom = {},
 		// Tests for existence of standard globals is to allow style-loader 
 		// to operate correctly into non-standard environments
 		// @see https://github.com/webpack-contrib/style-loader/issues/177
-        return window && document && document.all && !window.atob;
+		return window && document && document.all && !window.atob;
 	}),
 	getElement = (function(fn) {
 		var memo = {};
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
@FlyingDR Does JSDOM work 😛 ?

FlyingDR:
@michael-ciniawsky No idea about JSDOM, never had experience with it. Maybe @frodare can help with testing this change with JSDOM?

michael-ciniawsky:
@FylingDR Alright nevermind then 😛

frodare:
@FlyingDR this PR does fix the issue for our JSDOM testing environment.

michael-ciniawsky:
@FlyingDR Thx 😛

PR #208, opened 2017-04-04
test: usage with css-loader and composes

avaly:
Test cases: passing and failing

Yes, only tests.

**Summary**
Using `style-loader` and `css-loader` with local scoped identifiers and `composes` works as expected.

However when adding `css-loader`s `getLocalIdent` option, only some identifiers are properly passed through the custom `getLocalIdent`, while the identifiers from the composed files are still using the `localIdentName` option.

See the failing test `local scope, composing, custom getLocalIdent`

I submitted this PR here, because this is not reproducible when using only the `css-loader`.

### DIFF
--- a/test/basicTest.js
+++ b/test/basicTest.js
@@ -12,6 +12,12 @@ describe("basic tests", function() {
   var requiredCss = ".required { color: blue }",
     requiredCssTwo = ".requiredTwo { color: cyan }",
     localScopedCss = ":local(.className) { background: red; }",
+    localComposingCss = `
+      :local(.composingClass) {
+        composes: className from './localScoped.css';
+        color: blue;
+      }
+    `,
     requiredStyle = `<style type="text/css">${requiredCss}</style>`,
     existingStyle = "<style>.existing { color: yellow }</style>",
     checkValue = '<div class="check">check</div>',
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
@avaly Sry for the delay based on your description this sounds like a bug in `css-loader` instead?

michael-ciniawsky:
The CI is also failing but it seems this is intentional for demonstration?

avaly:
@michael-ciniawsky I'm not sure where the issue is between `style-loader` and `css-loader`. What's interesting is I can't reproduce this bug using only `css-loader`, only when throwing `style-loader` in the mix. I suspect it might be from how the options of `css-loader` are being handled by webpack when chaining the two loaders, since the `getLocalIdent` is a function, but not sure.

avaly:
@michael-ciniawsky Sorry for opening the PR here, but I tried adding this type of failing tests for the `css-loader` repo, but I could not replicate the bug. I will open a bug in that repo to make this visible there as well.

michael-ciniawsky:
I also need to take a deeper look at it first aswell 😛, yep please open an issue in `css-loader` so i can keep track of that

Is this PR for demonstration proposes only or subject to eventually merge in the future ? 🙃

avaly:
I was looking into the `css-loader` issues now, and I found https://github.com/webpack-contrib/css-loader/issues/389. It seems that the solution you suggested there (adding an `ident`) fixes this issue.

I've updated my branch and tests seem to pass. Feel free to merge it or not, though I feel it might be helpful for others who might run into this.

PR #219, opened 2017-04-21
refactor: use annotation comments for source maps (`options.sourceMap`)

### LINKED ISSUE #121: Blob links affects the styles later than JS execution
html:

```
  <style type="text/css">
  .test {
    background: red;
    width: 100px;
    height: 100px;
  }
  </style>
  <div class="test"></div>
```

css:

```
.test {
  display: none;
}
```

js:

```
require('./test.css');
$('.test').show();
```

In this case, if css-loader is used with ?sourceMap option, the css will be generated as Blobs. However, it seems that it will take sometime for blob link to load. So `$('.test').show();` executes before `display: none;` takes effect, and nothing shows up.
It's fine to print css directly on the page.

Is there any way to enable source-map without that problem?

mgol:
This is a rebase of PR #165 with minor modifications (e.g. a complete removal of `fixUrls.js`.

Fixes broken relative and protocol-relative url() use in stylesheets when
source maps are enabled. Chrome >=57 and Firefox >=50.1 support loading source
maps when sourceMapURL directive is dynamically added to a new or existing style
element in the DOM. This means that style-loader can simply add the styles and
sourceMapURL directive as text inside a style element and the source map will be
loaded by the browser. Given the simpler insertion method, the relative and
protocol-relative urls in stylesheets will not break.

Fixes #121
Ref #93

Bugfix + breaking change.

Not yet.

Yes.

**Summary**

Currently if source maps are enabled `style-loader` uses blob URLs instead of `<style>` tags with inline CSS. This makes the CSS loading asynchronous and causes subsequent JavaScript to often see the state from before applying the CSS, breaking apps (see #121). Another issue is breaking relative URLs (see #93).

Blobs were previously used to make source maps work but recent versions of major browsers have no problems with source maps working for inline `<style>` tags if a proper source map comment is appended.

Yes. The `convertToAbsoluteUrls` option is removed as it's no longer needed.

**Other information**
I can add some tests to verify the source map comment is appended and that it has correct mapping. Is anything else needed to land it?


mgol:
I set @john-bai as the author of the commit as almost the whole code is his. I can add tests in a separate commit.

michael-ciniawsky:
@mgol You mentioned Chrome && Firefox. Do you know the status in Safari/Edge ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "style-loader",
-  "version": "0.16.1",
+  "version": "0.16.2",
```
michael-ciniawsky: Please don't 😛, we use `standard-version` for release management across webpack-contrib and this is the responsibility of the respective maintainer

mgol: Oops. A bad merge, I didn't intend to mess with this field.

mgol: Fixed.

michael-ciniawsky: Yep no offense I thought so 😛

michael-ciniawsky:
@mgol Tests would be highly appreciated, but you mentioned is better do it in a separate PR which is ok as long as their will be one 😛 . Otherwise browser support and how strongly this affects current users is the elefant in the room :D

mgol:
@michael-ciniawsky I did some testing:

1. In Chrome 58 & Firefox 53 the approach from this PR works perfectly. One caveat is Chrome seems to apply the source map incorrectly if the inline style with a sourcemap is attached in the HTML statically instead of being created & appended via JS. This is not a problem for `style-loader`, though, as it will always do it via JS. I'll report a Chrome issue later.
2. In Safari 10.1 the blob approach works while the `<style>`-based one doesn't apply source maps.
3. In Edge 15 the `<style>`-based approach seems to work with some caveats (see the points below). The blob-based approach doesn't apply source maps. So this scenario is opposite to the Safari one. Caveats:
3.1. It seems to point to a place near the beginning of the file instead of the correct location... but at least it gets the file right.
3.2. It only works if you open DevTools *after* the style append happens. Otherwise you need to reopen DevTools to see the mapping.
4. In IE 11 source maps are never applied.

I still think the `<style>`-based approach is the way to go as the blob-based one causes very serious issues. The only issue is that will disable source map support in Safari. If we want to keep such support, this would require imtroducing yet another flag to let people opt into the blob approach. This would increase the complexity of `style-loader` code even more, though.

Let me know what you think.

michael-ciniawsky:
First thx very much for investigating on this futher 👍 , personally I would prefer a "Best used with Chrome/Firefox` :stuck_out_tongue: note/warning in the README instead of a bloated inconvenient 'ducktape' approach in `style-loader`, but that's not my call to make alone :).

michael-ciniawsky:
@mgol mind taking a look at #153 aswell  😛 ?

mgol:
@michael-ciniawsky does #153 influence this PR as well?

michael-ciniawsky:
@mgol I'm not 💯 about that fact

lydell:
I’m using this in a project, and it seems to work just fine. Sad to see that there are merge conflicts now, though!

mgol:
@lydell PR rebased.

@michael-ciniawsky how do we proceed further with this PR?

d3viant0ne:
@mgol  - We just need to plan exactly what is going into the next major release, i'll see that happens before the end of the weekend.

I'll tag a 0.x branch and we can start moving master to 1.x

mgol:
Cool; once you plan, please let me know what else needs to be done for this to land and I'll look into it. Thanks!

john-bai:
@mgol Thanks so much for rebasing and helping this PR land in a future version! I'm sorry I wasn't paying attention to my personal emails and missed all the updates. I kinda lost hope in the change landing but this is great to see it is close now 🎉

mgol:
There are conflicts again. I can rebase but I'd like to not keep rebasing every few weeks. :) What are the plans regarding landing this PR? @d3viant0ne has something stabilized since May?

michael-ciniawsky:
@mgol Could you rebase one more time and change the base branch `next` please? :)

mgol:
@michael-ciniawsky Done.

evilebottnawi:
@mgol maybe best create new PR?

mgol:
@evilebottnawi Why? It's already rebased. Do you want to have clearer conversation or something?

Note that @john-bai is still marked as the author of the commit in this PR as I based my PR on his one.

teriu:
Any progress on getting this PR merged in?

kevinzang:
Did this change get released? I was including it via the commit 45ba5c96d27d8454e56fce21322386a233e1488e, but starting today (2018-08-20) `yarn install` failed because 45ba5c96d27d8454e56fce21322386a233e1488e is no longer a valid commit (fatal: reference is not a tree). The merge commit (https://github.com/webpack-contrib/style-loader/pull/219/commits/8a179188ba549f0f49243f43c8cc97b660cc5c1b) seems to not be valid anymore either.

mdekalka:
Would like to add to @kevinzang comment that downloading the latest version `0.23.1` seems does not include this fix.

PR #232, opened 2017-05-15
fix: fallback for `document.querySelector`

andriichumak:

Bugfix. It's a fallback for older / non-standard devices, that do not support document.querySelector

No.

No.

**Summary**

Old / non-standard browsers would throw an error here. This fix will check if browser has querySelector and if it does not - fallback to putting styles to HEAD.

No.

**Other information**

This fix is important for e.g. HbbTv and SmartTv applications, that supposed to run on old devices with limited resources (so, too many polyfills is not welcome).

### DIFF
--- a/addStyles.js
+++ b/addStyles.js
@@ -27,7 +27,9 @@ var stylesInDom = {},
 			return memo[selector]
 		};
 	})(function (styleTarget) {
-		return document.querySelector(styleTarget)
+		return document && document.querySelector
+			? document.querySelector(styleTarget)
+			: document.getElementsByTagName('head')[0];
 	}),
 	singletonElement = null,
 	singletonCounter = 0,
--- a/addStyles.js
+++ b/addStyles.js
@@ -27,7 +27,7 @@ var stylesInDom = {},
 			return memo[selector]
 		};
 	})(function (styleTarget) {
-		return document && document.querySelector
+		return document.querySelector
 			? document.querySelector(styleTarget)
 			: document.getElementsByTagName('head')[0];
 	}),


Moln:
We have to support the old browsers.

evilebottnawi:
@Moln where?

michael-ciniawsky:
@andriichumak Do you use `style-loader` 'in production' ? http://caniuse.com/#search=querySelector supports all browsers >= IE11. I'm personally not in favour of this since `document.querySelector` has decent support cross browser support.

andriichumak:
@michael-ciniawsky Yes, it's used in production. I agree that it's fine to have querySelector for normal browsers. The problem occurs on Samsung HbbTv/SmartTv produced in 2012 and earlier and on some television set-top boxes. 5 years is very old for browsers, but in TV world it's still a pretty big market share. I'll have to support such devices for at least another 2 years.

I'm using my fork now. I have rather an edge case, so wouldn't mind if you decline PR. Nevertheless, it seems to be a painless fix. I could add a few lines to docs to make sure behavior is not confusing.

evilebottnawi:
@andriichumak can your provide example on your fork and explain what exactly has been done, perhaps this makes sense :+1:

andriichumak:
@evilebottnawi The only change is the one in this PR. Sorry for confusion.

bebraw:
@evilebottnawi To truly answer that, we should maintain a list of supported browsers somewhere on the main project most likely.

d3viant0ne:
@bebraw - last I asked, we support what webpack supports

bebraw:
@d3viant0ne Yup. So the question is, is that info somewhere public?

d3viant0ne:
@bebraw Not on the new ( or old ) docs site which is where I think users would expect to find that information. 

Probably wouldn't hurt to throw that & the supported Nodejs engines range somewhere under `https://webpack.js.org` perhaps `https://webpack.js.org/supportedversions/`

bebraw:
@d3viant0ne Ok. Better ask Tobias for an opinion. He should know the minimum versions. I think we dropped old IE a while back at least.

michael-ciniawsky:
A future proven compromise would be to pass e.g the selector method via options directly e.g `{ insert: document.getElementById('foo') }` etc, while not adding technical dept to the code base itself for edge cases

PR #244, opened 2017-05-29
chore(package): add `package-lock.json` (npm@5)

michael-ciniawsky:

 Chore

✅ 

**Summary**

- Add `package-lock.json` (` >= npm@5`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3665 @@
+{
+  "name": "style-loader",
+  "version": "0.18.0",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.3",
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@michael-ciniawsky It seems to me a bit early to do it

michael-ciniawsky:
Yep, not urgent at all

d3viant0ne:
This is actually an issue we need to discuss at the org level. i.e. Use npm || use yarn and then update `webpack-defaults` accordingly. Not really necessary, no prudent to keep both given I have seen dependency version related issues crop up with the differences with how they both work.

In either case, we should probably give `node@v8.0.0 / npm@5.0.0` a few weeks in the wild before we pick on or the other.

michael-ciniawsky:
https://github.com/webpack-contrib/organization/issues/13

PR #248, opened 2017-06-06
feat: add support for iframes (`options.insertInto`)

### LINKED ISSUE #213: insertInto for iframes
Feature

insertInto is limited to a string which is passed to document.querySelector

It would be nice if iframe support was included into insertInto as we often want to load css into an iframe for embedded widgets and the like.

tobias-zucali:

- add possibility to target an iframe with the `insertInto` setting
- select head in the iframes content document
- use try/catch to avoid breakting the code if permission is denied

Yes.

Yes.

**Summary**
Pull request for open issue **insertInto for iframes #213**

No.

**Other information**

Closes #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ By default, the style-loader appends `<style>` elements to the end of the style 
 ```
 
 ### `insertInto`
-By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else you can specify a CSS selector for that element here. If you target an iframe make sure you have sufficient access rights, the styles will be injected into the content document head.
+By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else you can specify a CSS selector for that element here. If you target an [IFrame](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement) make sure you have sufficient access rights, the styles will be injected into the content document head.
 You can also insert the styles into a [ShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot), e.g
 
 **webpack.config.js**
--- a/README.md
+++ b/README.md
@@ -286,7 +286,8 @@ By default, the style-loader appends `<style>` elements to the end of the style 
 ```
 
 ### `insertInto`
-By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else, e.g. into a [ShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot), you can specify a CSS selector for that element here, e.g
+By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else you can specify a CSS selector for that element here. If you target an iframe make sure you have sufficient access rights, the styles will be injected into the content document head.
+You can also insert the styles into a [ShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot), e.g
 
 **webpack.config.js**
 ```js
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -29,7 +29,7 @@ var getElement = (function (fn) {
 	return function(selector) {
 		if (typeof memo[selector] === "undefined") {
 			var styleTarget = fn.call(this, selector);
-			// Special case to return head of iframe insetead of iframe itself
+			// Special case to return head of iframe instead of iframe itself
 			try {
 				if (styleTarget instanceof window.HTMLIFrameElement) {
 					// This will throw an exception if access to iframe is blocked
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -30,14 +30,14 @@ var getElement = (function (fn) {
 		if (typeof memo[selector] === "undefined") {
 			var styleTarget = fn.call(this, selector);
 			// Special case to return head of iframe instead of iframe itself
-			try {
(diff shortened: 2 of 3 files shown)


test/basicTest.js
```diff
@@ -103,6 +104,17 @@ describe("basic tests", function() {
...
   }); // it insert into
 
+  it("insert into iframe", function(done) {
+    let selector = "iframe.iframeTarget";
+    styleLoaderOptions.insertInto = selector;
+
+    let expected = 'requiredStyle ';
+
+    runCompilerTest(expected, done, function() {
+      return this.document.querySelector(selector).contentDocument.head.innerHTML;
```
michael-ciniawsky: Why `this.document` and not just `document` ?

tobias-zucali: True, I can improve that ;-)

michael-ciniawsky: If it works without the `this` prefix, better ditch it :)

lib/addStyles.js
```diff
@@ -28,9 +28,19 @@ var getElement = (function (fn) {
 
 	return function(selector) {
 		if (typeof memo[selector] === "undefined") {
-			memo[selector] = fn.call(this, selector);
+			var styleTarget = fn.call(this, selector);
+			// Special case to return head of iframe insetead of iframe itself
```
bebraw: /s/insetead/instead

lib/addStyles.js
```diff
@@ -209,19 +219,19 @@ function addStyle (obj, options) {
...
-	    	// This allows running runtime transformations on the css.
-	    	obj.css = result;
-	    } else {
-	    	// If the transform function returns a falsy value, don't add this css.
-	    	// This allows conditional loading of css
-	    	return function() {
-	    		// noop
-	    	};
-	    }
+			result = options.transform(obj.css);
```
bebraw: I'm not sure if this formatting fix is needed. Any thoughts?

test/basicTest.js
tobias-zucali: `this` is needed. Seems that the test is run in node or so… but scope is explicitly set to window, that's why `this` works.

lib/addStyles.js
tobias-zucali: Was working too fast ;-) was mixed blanks with tabs, but the automatic conversion choose the wrong indent level. Should I fix the indentation level or revert it to mixed blanks/tabs?

michael-ciniawsky: I though it is github display madness with tabs tbh (I will send a style PR converting to spaces 😛). 
But in case it's not @tobias-zucali please use the indent the rest of the code currently uses.

michael-ciniawsky:
Fixes #213

lib/addStyles.js
```diff
@@ -209,19 +219,19 @@ function addStyle (obj, options) {
...
+			// If transform returns a value, use that instead of the original css.
+			// This allows running runtime transformations on the css.
+			obj.css = result;
+		} else {
+			// If the transform function returns a falsy value, don't add this css.
+			// This allows conditional loading of css
+			return function() {
+				// noop
+			};
+		}
```
d3viant0ne: Can you shift this block back, no reason for this to show up in the diff as there are not code changes less what looks like an auto-format.

lib/addStyles.js
```diff
@@ -28,9 +28,19 @@ var getElement = (function (fn) {
 
 	return function(selector) {
 		if (typeof memo[selector] === "undefined") {
-			memo[selector] = fn.call(this, selector);
+			var styleTarget = fn.call(this, selector);
+			// Special case to return head of iframe instead of iframe itself
+			try {
+				if (styleTarget instanceof window.HTMLIFrameElement) {
```
d3viant0ne: Do it not make more sense to check `instanceof` first? As it stands right now, we are going to execute that try/catch block every time regardless. When in reality, it only needs to execute if `styleTarget instanceof window.HTMLIFrameElement`

tobias-zucali: I was not sure in which environments style-loader can be run … is it possible that `window.HTMLIFrameElement` does not exist there?

d3viant0ne: My point was more that a truthy check is faster than executing a try catch iirc. Given the style loader chain already suffers from performance issues at scale, we need to be efficient when and wherever possible.

//cc @bebraw @michael-ciniawsky

michael-ciniawsky: Yep, `style-loader` should only support environments which expose/mimic a DOM Interface, the tests already use `JSDOM` which is the primary non-browser target `style-loader` supports atm. @tobias-zucali Could you try without the `try/catch` and `instanceof` check and see if it works ? Maybe add a comment with the current solution in case it heavily breaks other people's setups and we need a hotfix :). Otherwise good to go imho

README.md
```diff
@@ -286,7 +286,8 @@ By default, the style-loader appends `<style>` elements to the end of the style
 ```
 
 ### `insertInto`
-By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else, e.g. into a [ShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot), you can specify a CSS selector for that element here, e.g
+By default, the style-loader inserts the `<style>` elements into the `<head>` tag of the page. If you want the tags to be inserted somewhere else you can specify a CSS selector for that element here. If you target an iframe make sure you have sufficient access rights, the styles will be injected into the content document head.
```
michael-ciniawsky: `iframe` => `[IFrame](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement)`

Kavenon:
When will this feature be published in npm repo?

PR #250, opened 2017-06-22
fix(addStyles): use `typeof` to check for existence of `window` && `document` objects

bherila:

Fixes the following error which occurs during Webpack build by adding the check via the `typeof` operator. 

```
        return window && document && document.all && !window.atob;
               ^

ReferenceError: window is not defined
    at C:\Users\ucadmin\Documents\uc\admin\build\server.js:32858:9
```

n/a

n/a

**Summary**

Of note, other places in this file use the `typeof` operator, but curiously it is not done here.

No

**Other information**
n/a, it's a one-liner 😄

### DIFF
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -20,7 +20,7 @@ var isOldIE = memoize(function () {
 	// Tests for existence of standard globals is to allow style-loader
 	// to operate correctly into non-standard environments
 	// @see https://github.com/webpack-contrib/style-loader/issues/177
-	return window && document && document.all && !window.atob;
+	return (typeof window !== "undefined") && (typeof document !== "undefined") && document.all && !window.atob;
 });
 
 var getElement = (function (fn) {


bherila:
FYI this is in *addition* to https://github.com/webpack-contrib/style-loader/issues/177 which curiously is not enough 😦

michael-ciniawsky:
@bherila In which enviroment do you use the loader (JSDOM) ?

evilebottnawi:
@bherila can your provide use case, where your need this check? best - create test repo

bherila:
@michael-ciniawsky It was in my attempt to get https://github.com/barbar/vortigern working with bootstrap-loader (bootstrap 4). I have been attempting to get bootstrap 4 rendering with isomorphic-style-loader, and ideally have the bootstrap 4 critical path CSS analyzed and rendered server-side. As you can see from the vortigern site, there are a lot of packages being assembled together here :-\

Also, I would have shared the repo in which this reproduces, however it's for a client and I am not sure I can publish all the code on Github. I can try to make a minimal repro-repo if it would be helpful somehow.

veeramarni:
any update on merging this issue.

michael-ciniawsky:
@veeramarni Could you describe your 'use case' ? Why/Where are the current checks insufficient?  


> ⚠️ I'm inclined to close this PR within the next 48 hours in case no concrete reproduction is provided

veeramarni:
@michael-ciniawsky I still see those errors when running with SSR. In my case, we use webpack to create npm packages and we use these packages inn an application where we get `window undefined` error when we try to run with SSR. This could be the packaged webpack build has the `style-loader`.

michael-ciniawsky:
SSR isn't officially supported, this loader only works in the browser or browser-like env like JSDOM. 

>  we get window is not defined

Would this PR resolve your issue ?

veeramarni:
I will give a try with this PR and let you know.

PR #253, opened 2017-06-27
feat: support 'before' insertions (`options.insertAt`)

iancw:
Allow new styles to be inserted before other `<style>` or `<link>` elements that
exist on the page. This allows more precise ordering for complex style overrides
between multiple sources.

New feature

Yes

Yes

**Summary**

This is a new feature adding a third possible value for `insertAt`: `"before"`. A new option value `insertBefore` which contains query selector to select the element that the new styles should precede.

Complicated style situations where multiple `<style>` tags exist and must be inserted in specific order to allow style-overrides to work across sections require more precise placement than just `top` or `bottom`. This feature allows more precise selection of exactly where to insert a new `<style>` section relative to other existing sections.

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,8 +293,7 @@ A new `<style>` element can be inserted before a specific element by passing an 
   loader: 'style-loader'
   options: {
     insertAt: {
-        type: 'before',
-        selector: '#style-overrides'
+        before: '#style-overrides'
     }
   }
 }
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ By default, the style-loader appends `<style>` elements to the end of the style 
 }
 ```
 
-A new `<style>` element can be inserted before a specific element by passing an object with `type='before'`, e.g.
+A new `<style>` element can be inserted before a specific element by passing an object, e.g.
 
 **webpack.config.js**
 ```js
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`. 
 |**`transform`** |`{Function}`|`false`|Transform/Conditionally load CSS by passing a transform/condition function|
 |**`insertAt`**|`{String}`|`bottom`|Inserts `<style></style>` at the given position|
 |**`insertInto`**|`{String}`|`<head>`|Inserts `<style></style>` into the given position|
+|**`insertBefore`**|`{String}`|``|A query string specifying an element to insert new `<style>` before, required when `insertAt="before"`|
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`convertToAbsoluteUrls`**|`{Boolean}`|`false`|Coverts relative URLs to absolute urls, when source maps are enabled|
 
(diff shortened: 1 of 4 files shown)


lib/addStyles.js
```diff
@@ -170,6 +172,9 @@ function insertStyleElement (options, style) {
 		stylesInsertedAtTop.push(style);
 	} else if (options.insertAt === "bottom") {
 		target.appendChild(style);
+	} else if (options.insertAt === "before") {
+		var nextSibling = getElement(options.insertInto + " " + options.insertBefore);
+		target.insertBefore(style, nextSibling);
 	} else {
 		throw new Error("Invalid value for parameter 'insertAt'. Must be 'top' or 'bottom'.");
```
jbourjeli: Error message should include valid option 'before': 

> Invalid value for parameter 'insertAt'. Must be 'top', 'bottom' or 'before'.

michael-ciniawsky: Will this throw in the browser console or during build time? If the latter 

```js
this.emitError("Style Loader\n\n Invalid value for parameter 'insertAt' ('options.insertAt') found.\n Must be 'top' or 'bottom'.\n (https://github.com/webpack-contrib/style-loader#insertat)\n")
```

else 

```js
throw new Error("[Style Loader]\n\n Invalid value for parameter 'insertAt' ('options.insertAt') found.\n Must be 'top' or 'bottom'.\n (https://github.com/webpack-contrib/style-loader#insertat)\n")
```

README.md
```diff
@@ -139,6 +139,7 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`.
 |**`transform`** |`{Function}`|`false`|Transform/Conditionally load CSS by passing a transform/condition function|
 |**`insertAt`**|`{String}`|`bottom`|Inserts `<style></style>` at the given position|
 |**`insertInto`**|`{String}`|`<head>`|Inserts `<style></style>` into the given position|
+|**`insertBefore`**|`{String}`|``|A query string specifying an element to insert new `<style>` before, required when `insertAt="before"`|
```
michael-ciniawsky: I disagree with an extra top level option depending on other one. Possible to pass the an `{Object}` to `insertAt` instead `{ type: 'before',  $: '#id' }` (`$` maybe better named `selector`) [Needs discussion]

iancw: Yeah I wasn't completely sure the best way to approach this. The logic for `insertBefore` could also be triggered without requiring anything special in `insertAt`. In other words, if `insertBefore` has a non-empty string that behavior pre-empt the normal insertion logic triggered by `insertAt`.

I steered away from that because it seemed potentially confusing to have `insertBefore` pre-empt and essentially override the behavior of `insertAt`.

I'm also fine with making `insertAt` an object, but that becomes a breaking change. If you think that's best I can go with it, but it seems kind of heavy-handed for a small feature.

iancw: I see how it can be done in a backwards-compatible fashion. Pushed an implementation. Another option for the Object interface is `{ before: '#id'}`, which lets us avoid the ambiguous-sounding `type` key _and_ avoid choosing between `$` and `selector`.

michael-ciniawsky: Yeah an `{Object}` isn't optimal in terms of `type` consistency, but option dependenc isn't either, the interface you suggested is much better though :)

README.md
```diff
@@ -285,6 +285,20 @@ By default, the style-loader appends `<style>` elements to the end of the style
...
 
+A new `<style>` element can be inserted before a specific element by passing an object with `type='before'`, e.g.
+
+**webpack.config.js**
+```js
+{
+  loader: 'style-loader'
+  options: {
+    insertAt: {
+        before: '#style-overrides'
```
michael-ciniawsky: `#id || #name` (generic)

michael-ciniawsky: The options table  needs an `type` update aswell `{String} => {String\|Object}`  (⚠️ `\` [escaped])

michael-ciniawsky:
@webpack-contrib/org-maintainers please review :)

PR #264, opened 2017-09-08
feat: add option to enable/disable HMR (`options.hmr`)

### LINKED ISSUE #263: HMR code block should not be output for distribution, should accept an option

Bug

The output always has the following template (source code in index.js):
```
// Hot Module Replacement
if(false) {
	// When the styles change, update the <style> tags
	if(!content.locals) {
		module.hot.accept("!!../node_modules/some-npm-registry/postcss-loader/2.0.6/node_modules/postcss-loader/lib/index.js!./myComponent.pcss", function() {
			var newContent = require("!!../node_modules/some-npm-registry/postcss-loader/2.0.6/node_modules/postcss-loader/lib/index.js!./myComponent.pcss");
			if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];
			update(newContent);
		});
	}
	// When the module is disposed, remove the <style> tags
	module.hot.dispose(function() { update(); });
}
```

And this block is useless in prod because it will never be run with `if (false)` while HMR is off. But the problem is that the `"!!../node_modules/some-npm-registry/postcss-loader/2.0.6/node_modules/postcss-loader/lib/index.js!./myComponent.pcss"` is always being output in the dist file and breaks other bundling tools such as JSPM/SystemJS doing a static verification on the file path, the file path is only for local development and never been exported to dist.

**If the current behavior is a bug, please provide the steps to reproduce.**

The above HMR code block is always being generated in the final output.

The style-loader should introduce an option to not output this block if this flag is false, so that we could pass in the environmental flags. I would like to send a pull request for this and just wanted to make sure there's no side effect to just avoid outputting this block for production distribution when HMR is not needed.

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**

drawyan:
### Issues

- Fixes #263 

> edited by @michael-ciniawsky (Formatting && Issues)

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`. 
 
 |Name|Type|Default|Description|
 |:--:|:--:|:-----:|:----------|
+|**`hmr`**|`{Boolean}`|`true`|Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added (good for non local development/production)|
 |**`base`** |`{Number}`|`true`|Set module ID base (DLLPlugin)|
 |**`attrs`**|`{Object}`|`{}`|Add custom attrs to `<style></style>`|
 |**`transform`** |`{Function}`|`false`|Transform/Conditionally load CSS by passing a transform/condition function|
@@ -141,6 +142,21 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`. 
 |**`insertInto`**|`{String}`|`<head>`|Inserts `<style></style>` into the given position|
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`convertToAbsoluteUrls`**|`{Boolean}`|`false`|Converts relative URLs to absolute urls, when source maps are enabled|
+
+### `hmr`
+
+Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added.
+This could be used for non local development and production.
+
+**webpack.config.js**
+```js
+{
+  loader: 'style-loader'
+  options: {
+    hmr: false
+  }
+}
+```
 
 ### `base`
 
--- a/index.js
+++ b/index.js
@@ -17,6 +17,24 @@ module.exports.pitch = function (request) {
 
 	validateOptions(require('./options.json'), options, 'Style Loader')
 
+	options.hmr = typeof options.hmr === 'undefined' ? true : options.hmr;
+
+	var hmrCode = [
+		"// Hot Module Replacement",
+		"if(module.hot) {",
+		"	// When the styles change, update the <style> tags",
+		"	if(!content.locals) {",
+		"		module.hot.accept(" + loaderUtils.stringifyRequest(this, "!!" + request) + ", function() {",
+		"			var newContent = require(" + loaderUtils.stringifyRequest(this, "!!" + request) + ");",
+		"			if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];",
+		"			update(newContent);",
+		"		});",
+		"	}",
+		"	// When the module is disposed, remove the <style> tags",
+		"	module.hot.dispose(function() { update(); });",
+		"}"
+	].join("\n");
+
 	return [
 		"// style-loader: Adds some css to the DOM by adding a <style> tag",
 		"",
@@ -31,18 +49,6 @@ module.exports.pitch = function (request) {
 		"// add the styles to the DOM",
 		"var update = require(" + loaderUtils.stringifyRequest(this, "!" + path.join(__dirname, "lib", "addStyles.js")) + ")(content, options);",
 		"if(content.locals) module.exports = content.locals;",
-		"// Hot Module Replacement",
-		"if(module.hot) {",
-		"	// When the styles change, update the <style> tags",
-		"	if(!content.locals) {",
-		"		module.hot.accept(" + loaderUtils.stringifyRequest(this, "!!" + request) + ", function() {",
-		"			var newContent = require(" + loaderUtils.stringifyRequest(this, "!!" + request) + ");",
-		"			if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];",
-		"			update(newContent);",
-		"		});",
-		"	}",
-		"	// When the module is disposed, remove the <style> tags",
-		"	module.hot.dispose(function() { update(); });",
-		"}"
+		options.hmr ? hmrCode : ""
 	].join("\n");
 };
--- a/options.json
+++ b/options.json
@@ -2,6 +2,9 @@
 {
   "type": "object",
   "properties": {
+    "hmr": {
+      "type": "boolean"
+    },
     "base": {
       "type": "number"
     },
--- a/lib/addStyleUrl.js
+++ b/lib/addStyleUrl.js
@@ -18,6 +18,8 @@ module.exports = function addStyleUrl (url, options) {
 
 	options.attrs = typeof options.attrs === "object" ? options.attrs : {};
 
+	options.hmr = typeof options.hmr === 'undefined' ? true : options.hmr;
+
 	var link = document.createElement("link");
 
 	link.rel = "stylesheet";
@@ -30,7 +32,7 @@ module.exports = function addStyleUrl (url, options) {
 
 	head.appendChild(link);
 
-	if (module.hot) {
+	if (options.hmr && module.hot) {
 		return function(url) {
 			if(typeof url === "string") {
 				link.href = url;
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "jsdom": "^9.12.0",
     "memory-fs": "^0.4.1",
     "mocha": "^3.4.2",
+    "sinon": "^2.4.1",
     "standard-version": "^4.0.0",
     "webpack": "^2.6.1"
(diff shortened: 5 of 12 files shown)


options.json
```diff
@@ -22,6 +22,9 @@
     },
     "convertToAbsoluteUrls": {
       "type": "boolean"
+    },
+    "disableHMR": {
```
michael-ciniawsky: `disableHMR` => `hmr || hot` (with default `true`)

README.md
```diff
@@ -141,6 +141,7 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`.
 |**`insertInto`**|`{String}`|`<head>`|Inserts `<style></style>` into the given position|
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`convertToAbsoluteUrls`**|`{Boolean}`|`false`|Converts relative URLs to absolute urls, when source maps are enabled|
+|**`disableHMR`**|`{Boolean}`|`false`|Avoids outputting the unnecessary Hot Module Replacement code block, good for non local development distribution|
```
michael-ciniawsky:
```
|**`hmr || hot`**|`{Boolean}`|`true`|Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added (good for non local development/distribution)|
```

Add it as the first item in the options table please (above `base`)

README.md
```diff
@@ -344,6 +345,21 @@ If convertToAbsoluteUrls and sourceMaps are both enabled, relative urls will be
 }
 ```
 
+### `disableHMR`
```
michael-ciniawsky: `disableHMR` => `hmr || hot`

README.md
```diff
@@ -344,6 +345,21 @@ If convertToAbsoluteUrls and sourceMaps are both enabled, relative urls will be
 }
 ```
 
+### `disableHMR`
+
+When bundling for non development environment distribution, turn this option on to avoid outputting unnecessary Hot Module Replacement code block.
```
michael-ciniawsky: Reword this sentence please

michael-ciniawsky:
@drawyan Please also add a test for this

options.json
drawyan: Done

README.md
drawyan: Done

README.md
drawyan: Done

README.md
drawyan: Done

drawyan:
@michael-ciniawsky New patch set updated, sorry for the delay of the weekend. Many thanks for the review!

drawyan:
Sorry, accidentally closed it.

index.js
```diff
@@ -17,6 +17,24 @@ module.exports.pitch = function (request) {
 
 	validateOptions(require('./options.json'), options, 'Style Loader')
 
+	options.hot = typeof options.hmr === 'undefined' ? options.hot : options.hmr;
```
evilebottnawi: Can we use only one option? Example, i think `hmr` is better, because `hot` property is part of `hmr` (and can be misleading).

michael-ciniawsky:
Or `hot` depending on majority, but in any case just one property :)

drawyan:
OK, I'll go with `hmr`

README.md
```diff
@@ -142,6 +143,28 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`.
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`convertToAbsoluteUrls`**|`{Boolean}`|`false`|Converts relative URLs to absolute urls, when source maps are enabled|
 
+### `hmr || hot`
```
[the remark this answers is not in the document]
drawyan: Done

index.js
drawyan: Done

index.js
```diff
@@ -17,6 +17,24 @@ module.exports.pitch = function (request) {
...
+
+	var hmrBlock = [
```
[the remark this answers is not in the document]
drawyan: Done

index.js
```diff
@@ -31,18 +47,6 @@ module.exports.pitch = function (request) {
...
-		"		module.hot.accept(" + loaderUtils.stringifyRequest(this, "!!" + request) + ", function() {",
-		"			var newContent = require(" + loaderUtils.stringifyRequest(this, "!!" + request) + ");",
-		"			if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];",
-		"			update(newContent);",
-		"		});",
-		"	}",
-		"	// When the module is disposed, remove the <style> tags",
-		"	module.hot.dispose(function() { update(); });",
-		"}"
+		options.hmr ? hmr : ""
```
evilebottnawi: To avoid breaking changes set by default to `true`

README.md
```diff
@@ -142,6 +143,21 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`.
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`convertToAbsoluteUrls`**|`{Boolean}`|`false`|Converts relative URLs to absolute urls, when source maps are enabled|
 
+### `hmr`
+
+Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added.
+This could be used for non local development and distribution.
```
evilebottnawi: I think better to use `production`, `distribution` is misleading. `distribution` can be source and minified

index.js
```diff
@@ -17,6 +17,22 @@ module.exports.pitch = function (request) {
 
 	validateOptions(require('./options.json'), options, 'Style Loader')
 
+	var hmr = [
```
evilebottnawi: Let's return the name `hmrBlock`. `hmr` is option, hmrBlock` is code block for `hmr`.

index.js
michael-ciniawsky: Ohh 😲  good catch. Yep it **must** be `true` by default

index.js
michael-ciniawsky: The block doesn't add any semantic meaning, it already is an `{Array}`

evilebottnawi: @michael-ciniawsky `hmrCode` or `hmrSource`?

michael-ciniawsky: Why not just `hmr` as there isn't anything to distinguish from (atm). Only `options.hmr` (The Options) && `hmr` (The Code), The fact that it's a block (`{Array}` Type) and it is Source Code is already fulfilled 😛 , it just needs a name. `hmrCode` is also fine, but it's not really needed imho

evilebottnawi: @michael-ciniawsky I think better using different variable names, even if they are stored in different places, it makes the reading easier not to be misleading, also when using destructuring it helps to avoid accidentally overriding the variable

index.js
drawyan: This is already handled by the `options.json`'s `default` value

README.md
drawyan: Done

index.js
drawyan: I'll just leave as is for now

index.js
drawyan: Also my unit test covers the default case by not having any values for this option and it won't break because the ajv handles the `default` value from `options.json` correctly.

index.js
michael-ciniawsky: **if**, agreed. Still `const { code } = hmr` with `hmr` as the 'container/wrapper` **then** ;)

index.js
michael-ciniawsky: `options.hmr = options.hmr || true` (L19+) it shouldn't be mandatory to set this in `webpack.config.js` and it would be a breaking change aswell then. And please remove the `default: true` in case it causes a `ValidationError` for 

```js
{
   loader: 'style-loader',
   options: {}
},
{
   loader: 'style-loader',
   options: {
      other: option
   }
}
```

index.js
drawyan: hmm, `hmr` being a wrapper seems overkill here, but I do agree that it shouldn't be `hmr` to differentiate from the option's key, and this is why I put `hmrBlock` to indicate it's a code block. But as @michael-ciniawsky pointed out `block` being a bit confusing, I would say `hmrCode` is my pick. Updated.

index.js
drawyan: Done and updated.

index.js
```diff
@@ -16,6 +16,23 @@ module.exports.pitch = function (request) {
 	var options = loaderUtils.getOptions(this) || {};
 
 	validateOptions(require('./options.json'), options, 'Style Loader')
+	options.hmr = typeof options.hmr === 'undefined' ? true : options.hmr;
```
michael-ciniawsky: Please use `short-circuit` here (`options.hmr = options.hmr || true`)

michael-ciniawsky: and a space `\n`

drawyan: I don't understand how `options.hmr = options.hmr || true` is going to work. We'll never be able to set this flag to `false`, right?

drawyan:
@evilebottnawi I've updated other files, but not sure how to test them.

drawyan:
@evilebottnawi Mocked and updated, please review again. Thanks!

drawyan:
When will this be merged and released? My project is actually dependent on this. Thanks!

evilebottnawi:
I near future we release new version, thank you!

rsnara:
This is really awesome! :)

PR #270, opened 2017-10-26
feat: Parse default HMR option via npm environmental variable

drawyan:
This is to better address https://github.com/webpack-contrib/style-loader/issues/263

This will eliminate the dead code block for most of the packages that relying on `npm publish` to publish to npm repository.

Parse `options.hmr` value depends on the following condition:
  1. If `options.hmr` is a set value, return itself
  2. If `options.hmr` is missing with value `undefined`, try to parse the npm command used to
    invoke webpack via npm lifecycle event, to determine the default value:
      a) If invoked via npm `prepublish` or `prepublishOnly`, this is production distribution and
        default value for `options.hmr` to be false.
      b) If invoked via other npm scripts or non-npm commands, assuming this is local or dev env,
        default value for `options.hmr` to be true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`. 
 
 |Name|Type|Default|Description|
 |:--:|:--:|:-----:|:----------|
-|**`hmr`**|`{Boolean}`|`true`|Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added (good for non local development/production)|
+|**`hmr`**|`{Boolean}`|`true` for non-prod, `false` for prod|Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added. When in development, default value to be true, while in production, default to be false|
 |**`base`** |`{Number}`|`true`|Set module ID base (DLLPlugin)|
 |**`attrs`**|`{Object}`|`{}`|Add custom attrs to `<style></style>`|
 |**`transform`** |`{Function}`|`false`|Transform/Conditionally load CSS by passing a transform/condition function|
@@ -146,7 +146,11 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`. 
 ### `hmr`
 
 Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added.
-This could be used for non local development and production.
+If no specific value provided, the default value will be determined by the npm script that invoked webpack:
+  * If invoked via `npm preblish/prepublishOnly`, considered for production distribution, default to false.
+  * if invoked via other npm scripts or non-npm scripts, considered non-prod, default to true.
+
+You can always Provide a specific value to override the defaults.
 
 **webpack.config.js**
 ```js
--- a/lib/parseHmrOption.js
+++ b/lib/parseHmrOption.js
@@ -1,0 +1,36 @@
+// Detects the life cycle of npm for determination of whether it's production distribution
+var NPM_PREPUBLISH_EVENT_REGEX = /^prepublish(Only)?$/gi;
+
+/**
+ * Parse `options.hmr` value depends on the following condition:
+ *   1. If `options.hmr` is a set value, return itself
+ *   2. If `options.hmr` is missing with value `undefined`, try to parse the npm command used to
+ *     invoke webpack via npm lifecycle event, to determine the default value:
+ *       a) If invoked via npm `prepublish` or `prepublishOnly`, this is production distribution and
+ *         default value for `options.hmr` to be false.
+ *       b) If invoked via other npm scripts or non-npm commands, assuming this is local or dev env,
+ *         default value for `options.hmr` to be true.
+ * @param {Boolean|undefined} hmr - The original `options.hmr`
+ * @returns {Boolean}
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -146,7 +146,11 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`.
 ### `hmr`
 
 Enable/disable Hot Module Replacement (HMR), if disabled no HMR Code will be added.
-This could be used for non local development and production.
+If no specific value provided, the default value will be determined by the npm script that invoked webpack:
+  * If invoked via `npm preblish/prepublishOnly`, considered for production distribution, default to false.
```
onlywei: Typo here -- "preblish" should be "prepublish". Also "considered for production distribution" is grammatically strange. Consider just removing that phrase.

lib/parseHmrOption.js
```diff
@@ -0,0 +1,36 @@
...
+ *   2. If `options.hmr` is missing with value `undefined`, try to parse the npm command used to
+ *     invoke webpack via npm lifecycle event, to determine the default value:
+ *       a) If invoked via npm `prepublish` or `prepublishOnly`, this is production distribution and
+ *         default value for `options.hmr` to be false.
+ *       b) If invoked via other npm scripts or non-npm commands, assuming this is local or dev env,
+ *         default value for `options.hmr` to be true.
+ * @param {Boolean|undefined} hmr - The original `options.hmr`
+ * @returns {Boolean}
+ */
+module.exports = function(hmr) {
```
onlywei: Does this work even if the `package.json` is like this:

```json
{
  "scripts": {
    "build": "webpack",
    "prepublishOnly": "npm run build"
  }
}
```
?

drawyan: Hmmm, npm lifecycle event seems only picking up the immediate one, so in this case, it's `build`. I'll need to think about a better solution here.

d3viant0ne:
It's a loader & should not care about be aware of any particular package managers environment / lifecycle events.

It's options should come from the Webpack configuration & this particular feature can be achieved within a config. 

Making any particular loader aware of it's environment outside of what comes in via it's configured options is unfortunately a path we aren't interested in going down.

michael-ciniawsky:
I'm -1 on this approach in general as `npm_lifecycle`  detection is _way_ out-of-scope for this loader  and for webpack usage in general. What is the usecase ? How do you build your applications ? Is there a strong reason the provided option to disable HMR isn't sufficient/can't be used ? Setting `NODE_ENV` && detection of `webpack.HotModuleReplacementPlugin` usage would be an acceptable default, but detecting plugin usage is hardly possible atm and for the proposal to use `NODE_ENV` it simply doesn't work that way in the ecosystem either. Open an issue to triage HMR architectural/workflow changes in `webpack/webpack` please

**webpack.config.js**
```js
const env = process.env.NODE_ENV 

{
   loader: 'style-loader'
   options: {
      hmr: env === 'production' && false
   }
}
```

**package.json**
```json
{
   "scripts": {
      "prepublish":  "NODE_ENV=production webpack"
   }
}
```

drawyan:
The reason for this change is that we are not able to change how other published packages use webpack, a lot of published package just call `webpack` without any env flags for prod distribution, thus the `style-loader` will have this dead code block inside the distribution containing some dev static string path of css files, causing static path analyzers like JSPM/SystemJS to fail to bundle because those static path doesn't really exist.

Using npm lifecycle events isn't the ideal way, and I agree it's out of scope of the loader. I would love to ask if there are any other indicators or flags that we could use to eliminate the dead code block from being output.

There's another way of doing it: dynamically using `eval` function to insert or remove the code block in the run time when webpack's `module.hot` is available. But this is also not ideal due to the evil of `eval`.

michael-ciniawsky:
> The reason for this change is that we are not able to change how other published packages use webpack, a lot of published package just call webpack without any env flags for prod distribution, thus the style-loader will have this dead code block inside the distribution containing some dev static string path of css files, causing static path analyzers like JSPM/SystemJS to fail to bundle because those static path doesn't really exist.

How would this PR solve this problem ? It's annoying for sure, but without further action taken by the particular package authors this won't magically resolve. Unfortunately you will need to grind through the packages having this issue and open issues/pull requests to fix this with either way or do I miss something here? 

> There's another way of doing it: dynamically using eval function to insert or remove the code block in the run time when webpack's module.hot is available. But this is also not ideal due to the evil of eval.

Can you show a small code example how this would look like ?

```js
// new Function() === eval, but without global context via 'use strict' usage (safer)
if (module.hot) new Function([ 'module' ], `'use strict';\n\n${HMR_CODE}`)(module)
``` 

?

d3viant0ne:
I feel everyone sufficiently understands the reasoning. The fact remains, either suggested fix are admittedly less than optimal and not something that is going to find itself bing merged.

As @michael-ciniawsky said above `Open an issue to triage HMR architectural/workflow changes in webpack/webpack please`

The resolution for this particular issue has to start there.

drawyan:
Thanks @d3viant0ne and @michael-ciniawsky , I've opened up an issue here at webpack: https://github.com/webpack/webpack/issues/5891

Appreciated for the investigation and suggestions.

PR #273, opened 2017-11-10
fix(addStyles): allow `window` && `document` to not exist (isOldIE)

### LINKED ISSUE #272: isOldIE throws 'window is not defined' in non-browser environments
I'm using React server-side rendering in React.NET and compiling my app with `webpack`+`style-loader`+etc. `isOldIE` assumes `window` as a variable exists and can be falsy. The correct way to check for its existence would be `typeof` like in `lib/urls.js`.

```
ReferenceError: 'window' is undefined at Anonymous function (Script Document:700:1)
 at Anonymous function (Script Document:695:111)
 at module.exports (Script Document:702:23)
 at srcViewsRootStylesCss (Script Document:984:1)
 at __webpack_require__ (Script Document:50:9)
 ...
```

JoshuaKGoldberg:
Fixes #272.

Bugfix

No. Do you test this type of behavior?

**Summary**
Uses `typeof` checks so the function doesn't crash if `window` or `document` don't exist.

No.

### DIFF
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -20,7 +20,7 @@ var isOldIE = memoize(function () {
 	// Tests for existence of standard globals is to allow style-loader
 	// to operate correctly into non-standard environments
 	// @see https://github.com/webpack-contrib/style-loader/issues/177
-	return window && document && document.all && !window.atob;
+	return typeof window !== "undefined" && typeof document !== "undefined" && document.all && !window.atob;
 });
 
 var getElement = (function (fn) {


michael-ciniawsky:
Duplicate of #250

`style-loader`  requires a DOM and doesn't officially support SSR. Does 'SSR' work with your proposed fix ? Please provide more info about your setup. If the current behaviour just prevents some 'SSR-ish'  use case from working I can/will accept this change (#250)

JoshuaKGoldberg:
Ha, didn't see the dup. At least mine doesn't have unnecessary parenthesis...? 😄 

> Does 'SS' work with your proposed fix?

No, style-loader still crashes when it tries to use `document.createElement`. But I don't see any reason _not_ to allow this PR in: it doesn't hurt any behavior.

> setup && concrete use case

The project is compiled with Webpack into a single file that's use for client-side rendering (traditional use case). I'd like to take that file and dump it onto the server. This is the only blocker.

As a workaround, I forked our Webpack config file and replaced `style-loader` references with `ExtractTextPlugin` & similar to stub out the CSS... but it would be much cleaner to just use `style-loader` natively.

michael-ciniawsky:
kk, please use an proper alternative instead e.g [isomorphic-style-loader](https://github.com/kriasoft/isomorphic-style-loader). `style-loader` solely adds CSS via `<style>` to the DOM

indolent-developer:
i don't understand why are we just blatantly removing a PR. People are talking time to fix things which are breaking build which were working fine. This code line is wrong.
return window && document && document.all && !window.atob; whatever env we are running this.

The type check for window should be done better and that is what people are tying to do.
No where in code window checks are done in the way done here.
https://github.com/webpack-contrib/style-loader/search?utf8=%E2%9C%93&q=window&type=

michael-ciniawsky:
@indolent-developer I asked for a reproduction where these checks aren't actually working which isn't used for SSR and got no failing example until now. I'm more then open to merge this if it is really needed

PR #279, opened 2017-12-04
feat: Add the opportunity to use function to insertInto element

savelichalex:
This is new feature, to use function in `insertInto` like options, that add more flexibility to loader. I.e. my case - I want to insert css into shadow root, but this is impossible right now (because `::shadow` is deprecated)

Yes

Not sure about naming, need review first

**Summary**

Pass all existed test cases

**Other information**


test/basicTest.js
```diff
@@ -35,6 +35,13 @@ describe("basic tests", function() {
       "<iframe class='iframeTarget'/>",
       "</body>",
       "</html>"
+    ].join("\n"),
+    requiredJS = [
+      "var el = document.createElement('div');",
+      "el.id = \"test-shadow\";",
+      // "var shadow = el.attachShadow({ mode: 'open' })", // sadly shadow dom not working in jsdom
```
savelichalex: I really want to test with shadow root at first, but jsdom not working with shadow dom. Maybe some day we can use smth like headless-chrome for testing

lib/addStyles.js
```diff
@@ -23,7 +23,11 @@ var isOldIE = memoize(function () {
 	return window && document && document.all && !window.atob;
 });
 
-var getElement = (function (fn) {
+var getTargetFunction = function (target) {
```
michael-ciniawsky: `getTargetFunction` => `getTarget`

lib/addStyles.js
```diff
@@ -23,7 +23,11 @@ var isOldIE = memoize(function () {
...
+  return document.querySelector(target);
+};
+
+var getElementCreator = function (fn) {
```
michael-ciniawsky: `getElementCreator` => `getElement`

lib/addStyles.js
```diff
@@ -43,10 +47,9 @@ var getElement = (function (fn) {
 		}
 		return memo[selector]
 	};
-})(function (target) {
-	return document.querySelector(target)
-});
+};
 
+var getElement;
```
michael-ciniawsky: `getElement` => `element`

test/basicTest.js
michael-ciniawsky: [puppeteer](https://github.com/GoogleChrome/puppeteer)

index.js
```diff
@@ -44,8 +44,11 @@ module.exports.pitch = function (request) {
 		"// Prepare cssTransformation",
 		"var transform;",
 		options.transform ? "transform = require(" + loaderUtils.stringifyRequest(this, "!" + path.resolve(options.transform)) + ");" : "",
-		"var options = " + JSON.stringify(options),
-		"options.transform = transform",
+                "var insertIntoCallable;",
+                options.insertIntoCallable ? "insertIntoCallable = require(" + loaderUtils.stringifyRequest(this, "!" + path.resolve(options.insertIntoCallable)) + ")" : "",
```
michael-ciniawsky: This currently introduces a new options to the loader (`options.insertIntoCallable`) ?  I would be much in favor to resolve this via type checking of `options.insertInto` e.g `typeof options.InsertInto === 'function'` instead of adding yet another option to the loader. Passing a `{Function}` is a good idea I planned to reduce the `options`  of `style-loader` whenever possible in a similar way  for the next major release since most of current `options` are kind of `style-loader` specific 'weirdness'  due to technical debt and legacy support in the code.

lib/addStyles.js
```diff
@@ -67,7 +70,18 @@ module.exports = function(list, options) {
...
+        if (!options.insertInto) options.insertInto = "head";
+
+        // If provide function in options, then use it for resolve "head" element
+        // Usefull for Shadow Root style i.e
+        // {
+        // insertIntoCallable: "path/to/insertInto.js"
+        //}
+        // path/to/insertInto.js
+        // module.exports = function () { return document.querySelector("#foo").shadowRoot }
+        getElement = options.insertIntoCallable != null
```
michael-ciniawsky:
```js
var element = typeof options.insertInto === 'function' 
  ? getElement(options.insertInto) 
  : getElement(getTarget)
```

michael-ciniawsky: Another way maybe to pass the `options` to `getTarget` => 

```js
var getTarget = function (target, options) {
    return typeof options.insertInto === 'function' 
      ? options.insertInto() 
      : function (target) { return document.querySelector(target); }
}
```

Something like that

index.js
savelichalex: Yeah, first time I'm trying to use existing options but loader have validation by schema and this is validation check for string. At second there is should be a path to file. As we can't use function directly we can't check if with typeof operator and have 100% guarantee that given string is path to file

savelichalex:
@michael-ciniawsky I can't reply directly to comment for puppeteer 😄 I really like this idea, but I don't think that this PR a good place for changing test runner :) what do you think?

index.js
savelichalex: Another idea: what if use `.toString()` for function and then eval it. Not sure how it is safe

michael-ciniawsky: > Yeah, first time I'm trying to use existing options but loader have validation by schema and this is validation check for string.

This can be fixed as the `options` validation is able to handle that via `oneOf` && `typeof: 'function'` now (the dependency in `style-loader` likely needs an update). Simply remove the the check for `insertInto` while developing for now and if needed I will help/take over at this point then 

> As we can't use function directly we can't check it with typeof operator and have 100% guarantee that given string is path to file

**webpack.config.js**
```js
{
  insertInto: require('path/to/custom/targetFn') // {Object}
  insertInto () { // {Function}
    return document.querySelector('#foo')
 }
}
```

**index.js [L16+](https://github.com/savelichalex/style-loader/blob/7902ef3aafede882471fdb0b03d968541965b7e3/index.js#L16)**

```js
// handler (within the loader phase)
const insertInto = typeof options.insertInto === 'function' ? ...
...
// pass the result to code block
```

> Another idea: what if use .toString() for function and then eval (or inline) it. Not sure how it is safe

You can certainly try go in that direction since `webpack` will parse this code again so a function declaration as {String} should work (will be/should be 'evaluated' or parsed as `{Function}`)

michael-ciniawsky:
@savelichalex It's was just a suggestion not an requirement or the like :)

lib/addStyles.js
savelichalex: Not sure about this. Because in this place we create function nor element and then use it right there https://github.com/webpack-contrib/style-loader/blob/master/lib/addStyles.js#L154

index.js
savelichalex: > This can be fixed as the options validation is able to handle that via oneOf && typeof: 'function' now

Looks like `oneOf` in `ajv` not help, need to define new keyword

options.json
```diff
@@ -14,9 +14,6 @@
     "insertAt": {
       "type": ["string", "object"]
     },
-    "insertInto": {
-      "type": "string"
```
savelichalex: Looks like we need new keyword in ajv for this

michael-ciniawsky: Yep, a newer version of `schema-utils` is needed

index.js
```diff
@@ -35,6 +35,14 @@ module.exports.pitch = function (request) {
 		"}"
 	].join("\n");
 
+        var insertInto = "void 0";
```
michael-ciniawsky: Why `void 0` ?

index.js
```diff
@@ -35,6 +35,14 @@ module.exports.pitch = function (request) {
...
+        if (typeof options.insertInto === "function") {
+          insertInto = options.insertInto.toString();
+        }
+        if (typeof options.insertInto === "string") {
```
michael-ciniawsky: This check shouldn't be necessary as there isn't anything to handle within the pitching phase of the loader in case `options.insertInto` is a `{String}`

index.js
```diff
@@ -35,6 +35,14 @@ module.exports.pitch = function (request) {
 		"}"
 	].join("\n");
 
+        var insertInto = "void 0";
+        if (typeof options.insertInto === "function") {
```
michael-ciniawsky: 👍

lib/addStyles.js
```diff
@@ -23,12 +23,24 @@ var isOldIE = memoize(function () {
...
+  return document.querySelector(target);
+};
+
 var getElement = (function (fn) {
 	var memo = {};
 
-	return function(selector) {
-		if (typeof memo[selector] === "undefined") {
-			var styleTarget = fn.call(this, selector);
+	return function(insertInto) {
```
michael-ciniawsky: `insertInto` => `target`

lib/addStyles.js
```diff
@@ -23,12 +23,24 @@ var isOldIE = memoize(function () {
...
+                // If passing function in options, then use it for resolve "head" element.
+                // Usefull for Shadow Root style i.e
```
michael-ciniawsky: `Usefull` =>`Useful`

savelichalex:
@michael-ciniawsky Sorry for late response
> This also requires updates to the docs (README) then 😛
Agree :)

index.js
savelichalex: In this place https://github.com/savelichalex/style-loader/blob/fe58dbe66444852d9133a00d2d1883eb9036ff41/index.js#L55 I need some explicit value. This can be `undefined` or `null` of course)

index.js
savelichalex: I just want to make this check explicit, if you think that this is excess, I can delete this)

lib/addStyles.js
savelichalex: thanks!)

cthorner:
Would be nice with the shadow dom / iframe targeting to be able to load the styles to multiple targets. Maybe outside of the scope of this PR?

Thanks for adding this functionality.

savelichalex:
@cthorner Yeah, I think about it too! But unfortunately I don't know how to do it right. I'm also want to add styles not right after script evaluating. For me nice example of overriding runtime behavior is `svg-sprite-loader`

index.js
```diff
@@ -44,8 +52,10 @@ module.exports.pitch = function (request) {
 		"// Prepare cssTransformation",
 		"var transform;",
 		options.transform ? "transform = require(" + loaderUtils.stringifyRequest(this, "!" + path.resolve(options.transform)) + ");" : "",
-		"var options = " + JSON.stringify(options),
-		"options.transform = transform",
+                "var insertInto = " + insertInto + ";" ,
```
michael-ciniawsky: This is redundant with the reassignment on L57 ? 🙃

savelichalex:
@michael-ciniawsky I'll try to explain all this code :)
Why we need `var insertInto` and then assign to this variable instead of options? This is because if we just store function in options, then after `JSON.stringify` they quoted and in runtime will be just plain text nor function.
Why we need `typeof options.insertInto === "string"`? I found that without this check unspecified options become `"undefined"` and then in runtime throw an error, because of [this](https://github.com/webpack-contrib/style-loader/blob/master/lib/addStyles.js#L157).

michael-ciniawsky:
@savelichalex Please rebase once you have time, so I can finish this :)

jaesung2061:
What is the status on this PR? Can't you rebase yourself? @savelichalex @michael-ciniawsky

evilebottnawi:
@michael-ciniawsky need minor release :smile:

michael-ciniawsky:
Let's get #298 into `v0.20.0` aswell and then I cut a release

Issue #283, opened 2017-12-05
singleton, default is false, not true


qq20004604:
https://github.com/webpack-contrib/style-loader#singleton

> This option is on by default in IE9, which has strict limitations on the number of style tags allowed on a page. You can enable or disable it with the singleton option.

But in fact, singleton default is false, not true.

DEMO: https://github.com/qq20004604/webpack-study/tree/master/5%E3%80%81Loader/style_loader_demo

evilebottnawi:
@qq20004604 hm, you are right, let's fix it

evilebottnawi:
Done in https://github.com/webpack-contrib/style-loader/pull/285

PR #285, opened 2017-12-07
fix(addStyles): correctly check `singleton` behavior when `{Boolean}` (`options.singleton`)

evilebottnawi:

bugfix

yes

yes, we don't have `singleton` option in table options, i don't know how it is happens

**Summary**

When `singleton: true` - i want to use one `style` tag.
When `singleton: false` - i want to use multiple `style` tags.
When `singleton` not specify (i.e. `undefined`) - i want the loader himself decide to use single or multiple `style` tags.

Issue: https://github.com/webpack-contrib/style-loader/issues/283

No

**Other information**


evilebottnawi:
Btw be great to ensure `isOldIE` called or not using `sinon`, i try to found how to test this.

PR #288, opened 2017-12-29
feat: export `content`

cusspvz:
Exports the content so other loaders or apps could access to styles content directly.

No

No

**Summary**

I'm using styles content to fetch all styles hooked into rendered components at the server side so, instead of sending the CSS locations, the webapp renders with styles even before it is active.

I don't think so.

EDIT:
**Other information**
I just don't know if this should be provided optionally.

### DIFF
--- a/useable.js
+++ b/useable.js
@@ -45,6 +45,7 @@ module.exports.pitch = function (request) {
 		"var dispose;",
 		"var content = require(" + loaderUtils.stringifyRequest(this, "!!" + request) + ");",
 		"if(typeof content === 'string') content = [[module.id, content, '']];",
+		"exports.content = content;",
 		"if(content.locals) exports.locals = content.locals;",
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
@cusspvz I'm not sure `style-loader` should (re)export the  `content` CSS. The loaders sole purpose is to add a `<style></style` tag and SSR use cases are **not**  officially supported by this loader

Could you give describe your use case in more detail and give an example how the intended usage look like please ?

cusspvz:
> Could you give describe your use case in more detail and give an example how the intended usage looks like please ?

I have a implementation at the SSR that uses the `use/unuse` methods to flag which styles are activated on a page render. Then, I fetch all their style contents and send them into a preloaded `style` element, which is removed once the browser has completely rendered the page. This allows us to send a page as it was loaded and also improves the user experience.

So you don't think it should be exported?

PR #298, opened 2018-01-26
fix(index): enable HMR in case `locals` (`css-modules`) are unchanged

lydell:
fix/improvement

No, I couldn’t find that this type of thing has tests already. Nudge me in the right direction and I’ll add some!

I did make this test repo though to show that it works: https://github.com/lydell/css-hmr-bug/tree/fix

(The `master` branch shows the problem, while the `fix` branch shows the fix.)

Not relevant.

**Summary**
Fixes webpack-contrib/css-loader#669.

Previously, HMR was disabled if css-modules was enabled (in css-loader) and at least one class was present in the file. This was because JS files depend on the classes, and as such have to be reloaded when the CSS changes in case the classes have changed.

Now, HMR _is_ used if no classes change. Othewise the whole page is reloaded like before.

CSS HMR is awesome! css-modules are great, too. Now you can have both! :tada:

No breaking changes!


index.js
```diff
@@ -23,13 +23,34 @@ module.exports.pitch = function (request) {
 		"// Hot Module Replacement",
 		"if(module.hot) {",
 		"	// When the styles change, update the <style> tags",
-		"	if(!content.locals) {",
```
lydell: When I removed this check, I dedented the code so the diff looks a little bit larger than it is.

index.js
```diff
@@ -23,13 +23,34 @@ module.exports.pitch = function (request) {
...
+		"				localsSame = false;",
+		"			} else {",
+		"				localsSame = oldKeys.every(function(oldKey) {",
+		"					var oldValue = content.locals[oldKey];",
+		"					var newValue = newContent.locals[oldKey];",
+		"					return oldValue === newValue;",
+		"				});",
+		"			}",
+		"		}",
+		"		if(localsSame) {",
```
lydell: This comparison between old and new locals is the gist of this PR.

lydell:
Travis failed in Node.js 4.3 because of an unrelated issue?

```
$ yarn run test
error Couldn't find the binary yarn run test
```

The tests pass locally.

index.js
```diff
@@ -23,13 +23,34 @@ module.exports.pitch = function (request) {
...
 		"			update(newContent);",
-		"		});",
-		"	}",
+		"		} else {",
+		"			// This error is caught and not shown and causes a full reload.",
+		"			throw new Error('Aborting CSS HMR due to changed css-modules locals.');",
```
lydell: I looked through https://webpack.js.org/api/hot-module-replacement/ and tried to read the HMR source code in the webpack repo, and could not find any other way of aborting an `accept` other than throwing an error. It seems to work flawlessly, though!

lydell:
I’ve updated https://github.com/lydell/css-hmr-bug/tree/fix now to point to the correct git SHA for my style-loader fork, so it should work now.

index.js
```diff
@@ -23,13 +23,34 @@ module.exports.pitch = function (request) {
...
 		"if(module.hot) {",
 		"	// When the styles change, update the <style> tags",
-		"	if(!content.locals) {",
-		"		module.hot.accept(" + loaderUtils.stringifyRequest(this, "!!" + request) + ", function() {",
-		"			var newContent = require(" + loaderUtils.stringifyRequest(this, "!!" + request) + ");",
-		"			if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];",
+		"	module.hot.accept(" + loaderUtils.stringifyRequest(this, "!!" + request) + ", function() {",
+		"		var newContent = require(" + loaderUtils.stringifyRequest(this, "!!" + request) + ");",
+		"		if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];",
+		"		var localsSame = true;",
```
michael-ciniawsky: Is the name `locals` available ?

index.js
```diff
@@ -23,13 +23,34 @@ module.exports.pitch = function (request) {
...
+		"		if(!content.locals && newContent.locals) {",
+		"			localsSame = false;",
+		"		} else if(content.locals && !newContent.locals) {",
+		"			localsSame = false;",
+		"		} else if(content.locals && newContent.locals) {",
+		"			var oldKeys = Object.keys(content.locals);",
+		"			var newKeys = Object.keys(newContent.locals);",
+		"			if(oldKeys.length !== newKeys.length) {",
```
michael-ciniawsky:
```js
const length = Object.keys(content.locals).length - Object.keys(newContent.locals).length

if (length > 0) {
   ...
}
```

index.js
```diff
@@ -23,13 +23,34 @@ module.exports.pitch = function (request) {
...
+		"				localsSame = false;",
+		"			} else {",
+		"				localsSame = oldKeys.every(function(oldKey) {",
```
michael-ciniawsky:
```js
Object.keys(content.locals).every(function (key) {
   return content.locals[key] === newContent.locals[key]
})
```

index.js
michael-ciniawsky:
```js
if (!locals) throw new Error('...')
```

index.js
michael-ciniawsky: I will take a look into this, I think it's fine

index.js
lydell: Why? I find this harder to understand. And I use `oldKeys` a couple of lines down.

index.js
lydell: I guess. But isn’t that a worse name? `localsSame` answers the question “Are the locals the same?” while `locals` sounds like an object or array of local somethings.

lydell:
> This needs to be tested for the next branch aswell, where locals are named ESM Exports instead of an {Object} and if the approach taken here is upgradable

I know nothing about the `next` branch – what do I need to do more specifically?

index.js
lydell: `const length` – the length of what? There’s nothing with length 0 around.

lydell:
> Does this PR introduce a breaking change?

Hmm, I just realized `Object.keys` and `Array.prototype.every` aren’t supported in IE8 – is that a problem? Which browsers are supported?

index.js
michael-ciniawsky: Yep, sry I meant

```js
if (!length) {
  ...
}
```

index.js
michael-ciniawsky: The logic checks for identity of the locals (it's in the code), so if the name `locals` isn't already declared, `locals` is shorter and is imho clearer further below 

```js
// Do we have (new) locals => no HMR
if (!locals) throw new Error('..')
```

michael-ciniawsky:
Yeah that's probably a major then and would need be developed against the `next` branch

michael-ciniawsky:
What's the minimum IE version with support for `Object.keys` && `Array.every` ?

evilebottnawi:
@michael-ciniawsky we can accept this as bug here and release patch version, then do PR with this feature to next branch, after this will be resolved and merge

evilebottnawi:
Also I want to note that we have to have some time to support two versions (0.x.x and next) because, Not everyone can change the version quickly and without tests and regressions :smile:

michael-ciniawsky:
Not if this breaks the current browsers supported, which we need to figure out/ agreed upon a minimum version supported (if never specified somewhere)

michael-ciniawsky:
Agreed but this more a feature/refactor with eventual breaking changes for `master` then a (critical) bug

lydell:
IE9 is the lowest. I agree with @evilebottnawi , so I’ll get rid of `Object.keys` and `Array.prototype.every` so that we can merge this into `master`.

evilebottnawi:
@michael-ciniawsky it is looks as bug, because we don't have any note(s) about this don't works and i think it should be work as expected and says in docs.

index.js
lydell: Your comment is wrong.

- Do we have more locals than before? => no HMR
- Do we have fewer locals than before? => no HMR
- Do we have the same number of locals as before, but some changed? => no HMR
- Otherwise HMR

In other words:

- Are the locals the same? HMR
- Otherwise no HMR

index.js
lydell: Fixed in the latest commit.

lydell:
I've removed all ES5 usage now. The code became … _shorter!_

index.js
```diff
@@ -26,22 +26,15 @@ module.exports.pitch = function (request) {
...
-		"			var oldKeys = Object.keys(content.locals);",
-		"			var newKeys = Object.keys(newContent.locals);",
-		"			if(oldKeys.length !== newKeys.length) {",
-		"				localsSame = false;",
-		"			} else {",
-		"				localsSame = oldKeys.every(function(key) {",
-		"					return content.locals[key] === newContent.locals[key];",
-		"				});",
+		"		var localsSame = (function(a, b) {",
+		"			var key, numKeys = 0;",
```
michael-ciniawsky: `numKeys` => `idx`

index.js
```diff
@@ -26,22 +26,15 @@ module.exports.pitch = function (request) {
...
-		"		} else if(content.locals && newContent.locals) {",
-		"			var oldKeys = Object.keys(content.locals);",
-		"			var newKeys = Object.keys(newContent.locals);",
-		"			if(oldKeys.length !== newKeys.length) {",
-		"				localsSame = false;",
-		"			} else {",
-		"				localsSame = oldKeys.every(function(key) {",
-		"					return content.locals[key] === newContent.locals[key];",
-		"				});",
+		"		var localsSame = (function(a, b) {",
```
michael-ciniawsky: I still prefer `locals` here if available

index.js
```diff
@@ -26,22 +26,15 @@ module.exports.pitch = function (request) {
...
+		"			for(key in a) {",
+		"				if(!b || a[key] !== b[key]) return false",
+		"				numKeys++;",
 		"			}",
-		"		}",
+		"			for(key in b) numKeys--;",
+		"			return numKeys === 0;",
+		"		}(content.locals, newContent.locals));",
 		"		// This error is caught and not shown and causes a full reload.",
 		"		if(!localsSame) throw new Error('Aborting CSS HMR due to changed css-modules locals.')",
```
michael-ciniawsky: Missing `;` after `new Error('...')`

michael-ciniawsky:
@evilebottnawi Please review again, when time :)

index.js
```diff
@@ -23,13 +23,22 @@ module.exports.pitch = function (request) {
...
+		"			var key, idx = 0;",
+		"			for(key in a) {",
+		"				if(!b || a[key] !== b[key]) return false;",
+		"				idx++;",
+		"			}",
+		"			for(key in b) idx--;",
+		"			return idx === 0;",
+		"		}(content.locals, newContent.locals));",
+		"		// This error is caught and not shown and causes a full reload.",
+		"		if(!locals) throw new Error('Aborting CSS HMR due to changed css-modules locals.');",
```
evilebottnawi: Let's remove comment from source code (move as comment before this line), it is increase bundle in dev mode, Good work, thanks!

lydell: I see what you mean, but wouldn't that be inconsistent with all the other comments that are part of the output source code?

michael-ciniawsky: You're right, but the fact is that this code needs a few refactorings anyways in the future :)  @evilebottnawi is a good idea and **I will do it** for all the source comments before cutting a release. @evilebottnawi anything else of concern otherwise I would land this now and take care of the rest ?

evilebottnawi: @lydell oh yes, let's fix it in next PR

evilebottnawi: @michael-ciniawsky merge as is and fix it next PR

lydell:
Thank you both for taking the time to understand the issue, review and merge!

timbur:
You guys are awesome.  Thank you so much for fixing this!  ❤️ 🎉

GeorgeTaveras1231:
If I'm following this story correctly, it seems like now the style-loader _ALWAYS_ refreshes the page, when using CSS Modules. While the origin of the issue makes sense to me (JS files using CSS modules must be updated as well), this change will break HMR for apps that are 1. using CSS Modules, and 2. have HMR enabled for js files (such as using `react-hot-loader`). 

Am I missing something?

lydell:
>  it seems like now the style-loader _ALWAYS_ refreshes the page, when using CSS Modules

Nope, that was the previous behavior. The whole point of this PR was to _enable_ HMR for CSS modules – _when possible._

> react-hot-loader

I don't use react-hot-loader. Do you have an example where this PR broke something for you?

Issue #216, opened 2017-04-18
Perfect the custom attributes, Built-in component information, such as: name, version.


alex-mm:
## why?

I have many components.

![](https://gw.alicdn.com/tfscom/TB1xjK3QFXXXXXDaFXXXXXXXXXX.png)
 
I would like to distinguish between which component and the version number to use.

I found the `attrs`

```javascript
require('style-loader?{attrs:{id: "style-tag-id"}}!style.css');
```

But the information of the component can only be obtained in the compilation of each file.

so

I would like to make this change in 'style-loader/index.js'.

```javascript
	var findPackage = require('find-package');
        ...
	if(this.cacheable) this.cacheable();
	var query = loaderUtils.getOptions(this) || {};
	var pkg = findPackage(this.resourcePath) || {};
	if (query.attrs) {
		Object.keys(query.attrs).forEach(function(key) {
			query.attrs[key] = query.attrs[key].replace(/\[name\]/ig, pkg.name).replace(/\[version\]/ig, pkg.version);
		});
	}
        return ...
```

then

I use it like this

```
require('style-loader?{attrs:{component: "[name]@[version]"}}!style.css');
```

The result:

![](https://gw.alicdn.com/tfscom/TB1pLvwQFXXXXbRXFXXXXXXXXXX.png)

If it is acceptable ?

michael-ciniawsky:
@alex-mm What's the particular usecase for this in terms of development ? 😛

alex-mm:
@michael-ciniawsky 

It is mainly to help developers identify which components and versions.

When the dependent components are many, there are some common dependencies, especially when the dependent version is different, there will be style coverage problems, it is difficult to find which component is wrong version, there will also be repeated insert the situation, you are It is hard to see which component is repeated.

By adding the information on the 'attr', you can clearly find out the problem.

By making the private package, we've done it, it's really useful.

francoismassart:
I would appreciate it too. 

It is kind of frustrating to be able to add only "static values" to the attributes of the injected `<style>` elements...

michael-ciniawsky:
We could maybe include the filename, but I'm still not really convinced by the idea. If this is really wanted feel free/please open a PR for further discussion

budarin:
It might help you to remove stylesheet by id

kimown:
Really need this feature, like `webpackChunkName`, so we can distinguish source of styles.

alextrastero:
I'm using this:
```js
    options: {
[...]

PR #317, opened 2018-04-20
feat(index): add `data-url` attribute 

bukuta:

Feature

Yes

Yes

**Summary**

Add the support of dynamic attributes with placeholder as `file-loader` does,  supply some useful informations when debugging.

It may fix [the problem](https://github.com/webpack-contrib/style-loader/issues/216)

No breaking change

**Other information**
None

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ import style from './file.css'
 {
   test: /\.css$/,
   use: [
-    { loader: 'style-loader', options: { attrs: { id: 'id' } } }
+    { loader: 'style-loader', options: { attrs: { id: 'id' , 'x-from': '[path][name].[ext]'} } }
     { loader: 'css-loader' }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ import style from './file.css'
 {
   test: /\.css$/,
   use: [
-    { loader: 'style-loader', options: { attrs: { id: 'id' , 'x-from': '[path][name].[ext]'} } }
+    { loader: 'style-loader', options: { attrs: { id: 'id' , 'data-from': '[path][name].[ext]'} } }
     { loader: 'css-loader' }
   ]
 }
--- a/index.js
+++ b/index.js
@@ -13,15 +13,18 @@ module.exports.pitch = function (request) {
 	if (this.cacheable) this.cacheable();
 
 	var options = loaderUtils.getOptions(this) || {};
-	var context = options.context || this.rootContext || (this.options && this.options.context);
+
 	var attrs = {};
-	for(var key in options.attrs||{}){
-		var name = loaderUtils.interpolateName(this,options.attrs[key],{
-			context,
-			content:null,
-			regExp:options.regExp,
-		});
-		attrs[key]=name;
+	var dynamicAttributesMatcher = /\[\S+\]/;
+	var dynamicAttributesMatched = false;
+	var value;
+	for (var key in options.attrs || {}) {
+		value = options.attrs[key];
+		if(dynamicAttributesMatcher.test(value)){
+			dynamicAttributesMatched = true;
+			value = loaderUtils.interpolateName(this, value, options);
+		}
+		attrs[key] = value;
 	}
 
 	validateOptions(require('./options.json'), options, 'Style Loader')
@@ -93,7 +96,7 @@ module.exports.pitch = function (request) {
  		"",
 		"var options = " + JSON.stringify(options),
 		"",
-		"options.attrs = " + JSON.stringify(attrs),
+		dynamicAttributesMatched ? "options.attrs = " + JSON.stringify(attrs) : "",
 		"",
 		"options.transform = transform",
 		"options.insertInto = " + insertInto + ";",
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -12,7 +13,16 @@ module.exports = function () {};
 module.exports.pitch = function (request) {
 	if (this.cacheable) this.cacheable();
 
-	var options = loaderUtils.getOptions(this) || {};
+	var options = clone(loaderUtils.getOptions(this) || {});
```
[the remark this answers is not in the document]
bukuta: Ok.
`clone` was used because the recommend in `loader-utils`. 
Now it is removed, and the logic is splited

bukuta:
@evilebottnawi 

Some times we need some ways to find out 
- whether or not the styles processed with the right loaders and injected to dom
- the order  of the styles 

The differences can be distinct.

image
vs
image

evilebottnawi:
@bukuta Please accept `CLA`

bukuta:
@evilebottnawi  OK

package.json
```diff
@@ -15,6 +15,7 @@
     "options.json"
   ],
   "dependencies": {
+    "clone": "^1.0.4",
```
evilebottnawi: Please remove this

index.js
```diff
@@ -13,6 +13,16 @@ module.exports.pitch = function (request) {
 	if (this.cacheable) this.cacheable();
 
 	var options = loaderUtils.getOptions(this) || {};
+	var context = options.context || this.rootContext || (this.options && this.options.context);
+	var attrs = {};
+	for(var key in options.attrs||{}){
```
evilebottnawi: ` ` better `||`

index.js
```diff
@@ -13,6 +13,16 @@ module.exports.pitch = function (request) {
...
+		var name = loaderUtils.interpolateName(this,options.attrs[key],{
```
evilebottnawi: Can we add simple regex to detect `[here]`? To avoid perf problem?

package.json
bukuta: Removed

index.js
bukuta: Formatted

index.js
bukuta: I add a flag and a regexp

index.js
```diff
@@ -83,6 +96,8 @@ module.exports.pitch = function (request) {
  		"",
 		"var options = " + JSON.stringify(options),
 		"",
+		dynamicAttributesMatched ? "options.attrs = " + JSON.stringify(attrs) : "",
```
evilebottnawi: Just check `Object.keys(attrs).length > 0`

test/basic.test.js
```diff
@@ -240,6 +240,27 @@ describe("basic tests", function() {
...
+      rootDir + "main.js",
+      [
+        "var a = require('./style.css');"
+      ].join("\n")
+    );
+
+    // Run
+    let expected = [
+      existingStyle,
+      `<style x-from="style.css" type="text/css">${requiredCss}</style>`
```
evilebottnawi: Can we add tests around nested (i.e. with context option) to ensure path is right

README.md
```diff
@@ -210,7 +210,7 @@ import style from './file.css'
 {
   test: /\.css$/,
   use: [
-    { loader: 'style-loader', options: { attrs: { id: 'id' } } }
+    { loader: 'style-loader', options: { attrs: { id: 'id' , 'x-from': '[path][name].[ext]'} } }
```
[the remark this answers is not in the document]
evilebottnawi: :+1:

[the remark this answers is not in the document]
evilebottnawi: @bukuta let's use by default `data-from` in tests and README and we can merge this

README.md
```diff
@@ -210,7 +210,7 @@ import style from './file.css'
 {
   test: /\.css$/,
   use: [
-    { loader: 'style-loader', options: { attrs: { id: 'id' } } }
+    { loader: 'style-loader', options: { attrs: { id: 'id' , 'data-from': '[path][name].[ext]'} } }
```
michael-ciniawsky: `data-from` => `data-url`

index.js
```diff
@@ -14,6 +14,19 @@ module.exports.pitch = function (request) {
 
 	var options = loaderUtils.getOptions(this) || {};
 
+	var attrs = {};
```
michael-ciniawsky: Use the `request` or `this.resourcePath` and shorten it instead of interpolating, we already have the`url` here  

```js
var attrs = {
   'data-url': request || this.resourcePath
}
```

michael-ciniawsky:
Closing due to inactivity feel free to reopen once the requested chances are addressed :)

adriengcql:
Looks like it's almost done, can we reopen?

m-orsh:
Lets say I have 2 standalone apps. Each one needs to check whether the css libraries it needs have been loaded prior to loading them. If I load app 1, and this loads some css libs that app 2 also will use, without a method of identifying the css I have imported, how can I identify if the css has already been loaded? I think the only good way is with this feature.

Issue #306, opened 2018-03-06
__webpack_nonce__ support or alternative way of solving CSP?


madjam002:
I'm looking at using Webpack on a site which requires strict Content Security Policy headers.

I'm setting `window.__webpack_nonce__` but it doesn't seem to be adding a `nonce` attribute to `style` elements created by style-loader. Looking at webpack#3210 it only seems to be adding a nonce attribute to generated script elements? Is there some additional work required in style-loader to support nonce attributes?

I don't mind doing a PR, just wanted to make sure first though that I'm not missing something!

madjam002:
Just to add as well, I've noticed some discussion in #155 regarding CSP, but the way that custom attributes has been implemented there isn't suitable for nonce values as nonce values need to be secure cryptographically generated random strings for each page load and not hard coded into the build

jwbay:
The way aphrodite solves this is to look for an existing style tag before generating one. In their case something like `<style data-aphrodite>`. This effectively solves CSP because you can provide a style tag with a nonce already attached.

iamdavidfrancis:
Has there been any progress on this? I'm also working on a site with a strict CSP.
 It doesn't seem like it would be a huge PR to support __webpack_nonce__

evilebottnawi:
PR welcome :+1:

plondon:
I think this can be closed now since #319 has been merged and included in 0.22.0

plondon:
Also fyi @madjam002 you'll want to just set `__webpack_nonce__` and **not** `window.__webpack_nonce__`. Also do not try to do this with hot module reloading.

evilebottnawi:
@plondon can you send PR to README about this?

alshdavid:
What's the status with this?

Can I use style_loader with a nonce?
Is there anything special I need to do?
Will it read my window.__webpack_nonce__?
Will it work with dynamic imports?

plondon:
> What's the status with this?
> Resolved but confusing docs?
> Can I use style_loader with a nonce?
Yes
> Is there anything special I need to do?
Make sure the files where you set the nonce is one of the very first imports in the project:
[...]

PR #319, opened 2018-04-23
feat(addStyles): add support for `__webpack_nonce__`

iamdavidfrancis:
Feature to support `__webpack_nonce__` 

Yes, I added a test that verifies the nonce gets set on style tags. (Nonces will not help `link` tags as nonces are only used to allow inline styles.)

I wasn't sure where in the README I should put a note about supporting `__webpack_nonce__`, so I have omitted it for now. I can add it in if requested.

**Summary**
Currently if a site has a strict CSP that disallows `unsafe-inline` the rendered style tags from this loader will be ignored. Webpack supports adding a nonce via the `__webpack_nonce__` property. This change looks for that property and adds the nonce attribute to generated style tags.

More info here: #306 

This is only breaking if the package consumer was expecting the styles not to load, so it should not be an issue. If the consumer was already adding a nonce attribute via the options object, then that will take precedence over the `__webpack_nonce__` property, so it shouldn't break anyone doing it that way.

### DIFF
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -205,6 +205,13 @@ function createStyleElement (options) {
 		options.attrs.type = "text/css";
 	}
 
+	if(options.attrs.nonce === undefined) {
+		const nonce = getNonce();
+		if (nonce) {
+			options.attrs.nonce = nonce;
+		}
+	}
+
 	addAttrs(style, options.attrs);
 	insertStyleElement(options, style);
 
@@ -229,6 +236,14 @@ function addAttrs (el, attrs) {
 	Object.keys(attrs).forEach(function (key) {
 		el.setAttribute(key, attrs[key]);
 	});
+}
+
+function getNonce() {
+	if (typeof __webpack_nonce__ === 'undefined') {
+		return null;
+	}
+
+	return __webpack_nonce__;
 }
 
 function addStyle (obj, options) {
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -220,6 +220,27 @@ describe("basic tests", function() {
     runCompilerTest(expected, done);
   }); // it attrs
 
+  it("nonce", function(done) {
+    // Setup
+    const expectedNonce = "testNonce";
+
+    fs.writeFileSync(
+      rootDir + "main.js",
+      [
+        `__webpack_nonce__ = '${expectedNonce}'`,
+        "var a = require('./style.css');"
+      ].join("\n")
+    );
+
+    // Run
+    let expected = [
+      existingStyle,
+      `<style type="text/css" nonce="${expectedNonce}">${requiredCss}</style>`
+    ].join("\n");
+
+    runCompilerTest(expected, done);
+  }); // it attrs
+
   it("type attribute", function(done) {
     // Setup
     styleLoaderOptions.attrs = {type: 'text/less'};


evilebottnawi:
Also i think we should add this into https://github.com/webpack-contrib/style-loader/blob/master/lib/addStyleUrl.js

evilebottnawi:
@michael-ciniawsky let's leave this as is `attrs` is part of attribute tag, maybe change by default on `true` in major release

makeable:
Looking forward to this. It seems the previous workarounds of reusing nonced tags is no longer possible, so this is now the only legitimate fix for using style-loader with strict-dynamic CSP.

tobias-zucali:
Thanks for your work, I really need that!

I use webpack to load my app inside an iframe.
It works great for me as long as I use the [chrome csp tester plugin](https://chrome.google.com/webstore/detail/csp-tester/ehmipebdmhlmikaopdfoinmcjhhfadlf), but as soon as I deliver the csp header from the server it fails.
Strangely `__webpack_nonce__` cannot be accessed as a global, it needs to accessed via the window object.
I modified the following function of [addStyles.js](https://github.com/webpack-contrib/style-loader/blob/master/lib/addStyles.js#L204), it should never reach the log message as `getNonce()` should return `window.__webpack_nonce__` but in my case it logs the message.

Does the Content Security Policy restrict access to global scope? Any ideas?

``` js
function createStyleElement (options) {
	var style = document.createElement("style");

	if(options.attrs.type === undefined) {
		options.attrs.type = "text/css";
	}

	if(options.attrs.nonce === undefined) {
		const nonce = getNonce();
		if (nonce) {
			options.attrs.nonce = nonce;
		}
                 // MY EDIT START
                 else if (window.__webpack_nonce__) {
                         // This part should never be executed as 'getNonce()' should return the global variable… but in my case the message is logged 🤔
			console.log('`getNonce()` did not return anything but `window.__webpack_nonce__` is set!')
			options.attrs.nonce = window.__webpack_nonce__;
		}
                 // MY EDIT END
	}

	addAttrs(style, options.attrs);
	insertStyleElement(options, style);

	return style;
}

/* … */

function getNonce() {
	if (typeof __webpack_nonce__ === 'undefined') {
		return null;
	}

	return __webpack_nonce__;
}
```

tobias-zucali:
We spent more time on this issue and it works now without changing `style-loader`.

It is all about how to set `__webpack_nonce__ `. It is not enough to set `window.__webpack_nonce__ `.

This way it works:

webpack.config.js:
``` js
module.exports = {
  entry: {
    main: 'my/entry/index.js'
  }
}
```

my/entry/index.js
``` js
__webpack_nonce__ = /* getting the nonce from somewhere */
// will compile to `__webpack_require__.nc = /* getting the nonce from somewhere */`

// you must not use import here as it is hoisted and executed first (https://github.com/webpack/webpack/issues/2776)
require('./theRealEntry.js')
```

./theRealEntry.js
``` js
// `__webpack_nonce__` (compiled to `__webpack_require__.nc`is available in the style loader now
import('./styles.css')

// …
```

tobias-zucali:
... and it does not work at all if the HotModuleReplacementPlugin is active as different scopes are used …

plondon:
@tobias-zucali thank you so much for your comment on HMR. That was the issue I was having

evilebottnawi:
Somebody can create new issue and minimum reproducible test repo?

plondon:
@evilebottnawi I think its an issue with HMR not style-loader (or any other loader/project) that uses webpack_nonce. For example, I had the same issue with styled-components until I turned of HMR

bwhitty:
> I modified the following function of addStyles.js, it should never reach the log message as getNonce() should return window.__webpack_nonce__ but in my case it logs the message.

I am seeing this exact same thing described above where when actually returning the nonce the value is undefined. See the following recording:

csp-__webpack_nonce__

Note that the global is there, but upon return its value is `undefined`!

CSP I'm sending looks like:
```
default-src 'self' <couple of domains>; script-src 'self' <couple of domains> 'nonce-a78ee187-3bf1-46cd-b408-88a9f79e1806' 'strict-dynamic'; style-src 'self' <couple of domains>; img-src 'self' <couple of domains> data:; object-src 'self' <couple of domains>; media-src 'self' <couple of domains>; connect-src 'self' <couple of domains>
```

Modifying `addStyles.js` to be `global.__webpack_nonce__` makes it return the value, but when `insertStyleElement` is called I get the same CSP violation. From what I can tell, it's correctly setting the `nonce` attribute on the `style` object before attempting to append it to the DOM...

csp-__webpack_nonce__-global

Any ideas here?

plondon:
What does your create-nonce file look like?

plondon:
See our working example in this repo:

./create-nonce sets `__webpack_nonce__` from window.NONCE (passed from server) https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4-frontend/src/create-nonce.js

loaded in index.js
https://github.com/blockchain/blockchain-wallet-v4-frontend/blob/development/packages/blockchain-wallet-v4-frontend/src/index.js

bwhitty:
My client bundle index file essentially looks like this (modified to use `require(app)` to follow your instructions above regarding es module hoisting)

```jsx
... snip ...

const {
  csrfToken,
  nonce,
  reduxState,
  staticAssetsHost,
} = window.config;

// @see https://webpack.js.org/guides/csp/
// @see https://github.com/webpack-contrib/style-loader/pull/319
__webpack_nonce__ = nonce; // eslint-disable-line camelcase, no-undef

... snip ...

const App = require('app/client/components/app').default;

window.main = async function main() {
  // Wait for loadable to have retrieved all of its chunks
  // See: https://github.com/thejameskyle/react-loadable#waiting-to-render-on-the-client-until-all-the-bundles-are-loaded
  await Loadable.preloadReady();

  ReactDOM.hydrate(
    <ReduxProvider store={store}>
      <BrowserRouter>
        <Provider theme="light">
          <App />
        </Provider>
      </BrowserRouter>
    </ReduxProvider>,
    document.getElementById('app'),
  );
};

```

bwhitty:
config looks like (is this the hmr you're referring to?)
```
        test: /\.css$/,
        use: [
          {
            loader: 'style-loader',
            options: {
              hmr: false,
            },
          },
          'css-loader',
        ],
```

plondon:
Can you share the whole webpack config?

plondon:
And you shouldn't need to do this:

> Modifying addStyles.js to be global.__webpack_nonce__

plondon:
Nothing stands out in the webpack config.

Maybe the issue is you are not importing the nonce but setting it in App.js (or something similar). See comment here: https://github.com/styled-components/styled-components/issues/887#issuecomment-376536137

Issue #321, opened 2018-05-02
Transform not applied during HMR update


labriola:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
on HMR, defined transform functions are not applied to the reloaded content

**If the current behavior is a bug, please provide the steps to reproduce.**
Defined a style-loader, use webpack dev server, and make a change to the css. The css will be injected raw and never passed through the transform function.

**What is the expected behavior?**
I think that the transform should likely be applied to subsequent replacements else the first and nth load have potentially different CSS

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**
Webpack 4.x. Core issue seems to be that only addStyle applies the transform and that updateStyle does not.

evilebottnawi:
option transform was removed without replacement

PR #322, opened 2018-05-02
Possible naive fix for issue #321

labriola:
Potentially naive fix for issue #321 

No tests added as the issue only manifests under HMR and didn't see a correlated way to add those in the transform folder.

I don't believe it's relevant at this time

**Summary**
Ensures that transform function, when defined, is applied to both initial load and updates via HMR

Potentially naive fix for issue #321 

The only possibility is if someone was expecting an initial load and subsequent reload of a CSS file to be different. I don't know of a use case.

### DIFF
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -77,7 +77,7 @@ module.exports = function(list, options) {
 	if (!options.singleton && typeof options.singleton !== "boolean") options.singleton = isOldIE();
 
 	// By default, add <style> tags to the <head> element
-        if (!options.insertInto) options.insertInto = "head";
+	if (!options.insertInto) options.insertInto = "head";
 
 	// By default, add <style> tags to the bottom of the target
 	if (!options.insertAt) options.insertAt = "bottom";
@@ -243,11 +243,17 @@ function addStyle (obj, options) {
 	    	// This allows running runtime transformations on the css.
 	    	obj.css = result;
 	    } else {
-	    	// If the transform function returns a falsy value, don't add this css.
-	    	// This allows conditional loading of css
-	    	return function() {
-	    		// noop
-	    	};
+			// If the transform function returns a falsy value, don't add this now css.
+			// This allows conditional loading of css
+
+			// However, there is actually another bug here.
+			// There is no guarantee that the transform is a pure function so
+			// it may returns truthy when it updates later and we should potentially
+			// add the style at that time. However, there is a lot of refactoring
+			// required to adapt to that need
+			return function () {
+				// noop
+			};
 	    }
 	}
 
@@ -288,16 +294,16 @@ function addStyle (obj, options) {
 		var result;
 
 		if (newObj) {
-			if (options && options.transform && obj.css) {
-				result = options.transform(obj.css);
+			if (options && options.transform && newObj.css) {
+				result = options.transform(newObj.css);
 				if (result) {
-					newObj.css = result; 
+					newObj.css = result;
 				} else {
 					remove();
 					return;
 				}
 			}
-			
+
 			if (
 				newObj.css === obj.css &&
 				newObj.media === obj.media &&
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -123,7 +123,7 @@ function addStylesToDom (styles, options) {
 			domStyle.refs++;
 
 			for(var j = 0; j < domStyle.parts.length; j++) {
-				domStyle.parts[j](item.parts[j]);
+				domStyle.parts[j](item.parts[j], options);
 			}
 
 			for(; j < item.parts.length; j++) {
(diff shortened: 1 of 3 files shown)


labriola:
This branch now has tests.... Unfortunately without major refactoring there is no way to write the tests to the degree I would like. There is some package level memoization and caching that make things more stateful than they should be across tests. Also seeing a new issue, filed as #324 that was not possible to fix at this time.

test/addStyles.test.js
```diff
@@ -0,0 +1,154 @@
...
+      jsdomHtml,
+      undefined,
+      0,
+      done,
+      applyUpdate
+    )
+   });
+
+  // See issue #324, place holder for a future test once that issue is fixed
+  it.skip("should load the css if the transform function returns data during HMR");
```
evilebottnawi: What actually we need to solve this problem?

labriola: The discussion we had over on issue #324 there is quite a bit of refactoring. I can remove the test if you want but I am not in a position to spend more time on this at present

lib/addStyles.js
```diff
@@ -107,7 +107,6 @@ module.exports = function(list, options) {
 
 			if(domStyle.refs === 0) {
 				for (var j = 0; j < domStyle.parts.length; j++) domStyle.parts[j]();
-
 				delete stylesInDom[domStyle.id];
```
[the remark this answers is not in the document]
labriola: It was unintentional, I had a debugger statement there and, when I removed it, the newline was removed too. Sorry. 

The tests I have right now cover the changes but were extremely difficult to get in place with this module. If you are interested I can try to help refactor addstyles in the future, but I can't imagine adding more tests to this PR at the moment

test/addStyles.test.js
evilebottnawi: @labriola you can left this test, it is normal, we should add more tests to increase coverage

labriola:
Requests for fix to line break and tab resolved.

labriola:
I also just fixed a mistake which was causing the code coverage to drop so much.

evilebottnawi:
@labriola hm strange `(-26.16%)` :confused:

evilebottnawi:
In short: now transform works always as expected, exclude for hmr right?

labriola:
@evilebottnawi the code coverage only appeared to drop as there were never any tests for addStyles before. Now that there are, its counting addStyle as part of the total. So, we have increased test coverage but decreased the reported percentage.

In short, transform works all the time now, including during hmr.

There is one edge case left during hmr . If the first time the module is processed the transform returns falsey, the module will never be added. This is an edge case though and probably not likely, I just like to be thorough so I documented it as  issues #324

labriola:
I added the test for issue #324 as requested. It is set to skip() now but I will try to address it in a separate PR when possible. For now, it will demonstrate the issue.

evilebottnawi:
@labriola I have already requested an additional review from other contributors (just wait, i also don't love long merges, but it is package uses very very very many people and it is require more times on review ), do not worry. The only thing I'm afraid is that they can reject because due hmr

mauricecruz:
Resurfacing this PR. Running across the same issue and hoping I don't need to fork either.

Please let me know if there's anything I can do to move this along.

test/addStyles.test.js
```diff
@@ -0,0 +1,193 @@
...
+      jsdomHtml,
+      undefined,
+      0,
+      done,
+      applyUpdate
+    )
+   });
+
+  // See issue #324, place holder for a future test once that issue is fixed
+  it.skip("should load css if the transform function intially returns false but returns data during subsequent HMR", function(done) {
```
evilebottnawi: We should fix this problem here

labriola:
//cc @evilebottnawi Quick reminder. This PR is _not_ waiting on me. 

I increased the overall test coverage as my PR covers my code completely. You are seeing a decline in the project test coverage _percentage_ as my PR _exposes_ the fact that this project is completely missing test coverage of critical modules. Please see comment thread ending on May 4, 2018

If you would like, I will happily resubmit my PR, and remove all of my test. The net result will be that your test coverage will go up dramatically. However, I am not adding test coverage to unrelated code that I did not modify in this particular PR. It makes the purpose of this PR unclear and I would consider it an anti-pattern.

evilebottnawi:
@labriola thanks for clarify, i will review this in near future

evilebottnawi:
Sorry for delay, and sorry we can't merge this because transform was removed without replacement

labriola:
Amusing end to a 1.25 year saga to get a PR merged. At least we have reached a resolution

evilebottnawi:
@labriola it is open source and sometimes do not have enough time on all PRs/issues, sorry, we prepare style-loader/css-loader to built-in in webpack, so many option was removed

labriola:
@evilebottnawi I really do understand but that is a self-perpetuating problem. Had we been able to get this PR merged, I would have fixed a dozen more issues and been happy to help more. Unfortunately, if projects discourage (actively or passively) people from contributing, they end up with less and less time as projects grow. I am not criticizing you personally and I really do empathize with time constraints, I just see it as unfortunate. That said, best of luck, I will look at what comes next in webpack and try to fix any issues there.

Issue #332, opened 2018-07-10
[Bug Report] Raise error when export default used instead of module.exports


akki-jat:
**What is the current behavior?**
Raise error when export default used instead of module.exports.

```
export default function (css) {
  if (/*@cc_on!@*/false || !!window.document.documentMode) {
    css = css.replace(/var\(--primary-color\)/g, sharedService.primaryColor);
    css = css.replace(/var\(--secondary-color\)/g, sharedService.secondaryColor);
    css = css.replace(/var\(--text-color\)/g, sharedService.textColor);
    css = css.replace(/var\(--grid-col\)/g, sharedService.colsPercentage);
  }

  return css;
} 
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Just use export default instead of module.exports

**What is the expected behavior?**
Should not raise error

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**
webpack version : 3.6.0
Node.js version : 8.11.2
OS : Windows 7

**Error Screenshots**
![error_img](https://user-images.githubusercontent.com/26685162/42509161-163bfd1a-8419-11e8-82e9-3cdfc0f03543.png)

![error_img_2](https://user-images.githubusercontent.com/26685162/42509163-191d835a-8419-11e8-8efc-13aabc9c85c0.png)

![error_img_3](https://user-images.githubusercontent.com/26685162/42509169-1b5fc740-8419-11e8-994a-86ffe1a44372.png)

**Currently using this solution**
```
result = typeof options.transform === 'function'
           ? options.transform(obj.css) 
	   : Object.values(options.transform)[0](obj.css);
```

sokra:
Looks like a configuration problem. What's your loader configuration?

akki-jat:
Hi @sokra, here is my loader configuraion

```
{
    test: /\.css$/,
    use: [
        {
            loader: 'style-loader',
            options: {
                transform: './transform.js'
             }
          },
          {  loader: 'css-loader'  }
    ]
}
```

akki-jat:
@evilebottnawi and @sokra, any response on this?

akki-jat:
@evilebottnawi here is the repo to reproduce the bug.
[https://github.com/akki-jat/style-loader-bug--332](https://github.com/akki-jat/style-loader-bug--332)

evilebottnawi:
@akki-jat right now not, very busy, feel free to investigate and send fix

akki-jat:
[...]

PR #333, opened 2018-07-10
fix(addStyles): support exports of transpiled transforms (`options.transform`)

akki-jat:

bugfix

No

Solves #332 

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ A `transform` is a function that can modify the css just before it is loaded int
 This function will be called on the css that is about to be loaded and the return value of the function will be loaded into the page instead of the original css.
 If the return value of the `transform` function is falsy, the css will not be loaded into the page at all.
 
-> Warning: In case you are using ES Module syntax in tranform.js then you have to transpile it otherwise it will throw error.
+> :Warning: In case you are using ES Module syntax in `tranform.js` then, you **need to transpile** it or otherwise it will throw an `{Error}`.
 
 **webpack.config.js**
 ```js
--- a/README.md
+++ b/README.md
@@ -246,6 +246,8 @@ A `transform` is a function that can modify the css just before it is loaded int
 This function will be called on the css that is about to be loaded and the return value of the function will be loaded into the page instead of the original css.
 If the return value of the `transform` function is falsy, the css will not be loaded into the page at all.
 
+> Warning: In case you are using ES Module syntax in tranform.js then you have to transpile it otherwise it will throw error.
+
 **webpack.config.js**
 ```js
 {
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -254,7 +254,9 @@ function addStyle (obj, options) {
 
 	// If a transform function was defined, run it on the css
 	if (options.transform && obj.css) {
-	    result = options.transform(obj.css);
+	    result = typeof options.transform === 'function'
+		 ? options.transform(obj.css) 
+		 : options.transform[Object.keys(options.transform)[0]](obj.css);
 
 	    if (result) {
 	    	// If transform returns a value, use that instead of the original css.
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -256,7 +256,7 @@ function addStyle (obj, options) {
 	if (options.transform && obj.css) {
 	    result = typeof options.transform === 'function'
 		 ? options.transform(obj.css) 
-		 : options.transform[Object.keys(options.transform)[0]](obj.css);
+		 : options.transform.default(obj.css);
 
 	    if (result) {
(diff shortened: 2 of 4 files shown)


evilebottnawi:
Please accept CLA too

lib/addStyles.js
```diff
@@ -254,7 +254,9 @@ function addStyle (obj, options) {
 
 	// If a transform function was defined, run it on the css
 	if (options.transform && obj.css) {
-	    result = options.transform(obj.css);
+	    result = typeof options.transform === 'function'
+		 ? options.transform(obj.css) 
+		 : Object.values(options.transform)[0](obj.css);
```
evilebottnawi: One question, why `0`? Maybe better `default`?

evilebottnawi:
Error: `Error: Uncaught [TypeError: Object.values is not a function]`, `Object.values` is not supported in old browsers

akki-jat:
Yeah, you are right. So then we can use `Object.keys` or you suggest simply `options.transform.default(obj.css);`

akki-jat:
@evilebottnawi I replaced Object.values with Object.keys to support older browser.
But if you still prefer `options.transform.default(obj.css);` then tell me I will add that.

akki-jat:
@evilebottnawi can you now take a look?

akki-jat:
@michael-ciniawsky it's throw error: **transform is not a function** when **export default or exports.default** is used instead of **module.exports**.
Because when we use export default or exports.default, it returns the transform as object **(not as a function)** and actual transform function wrapped inside its **default** property

If want to check, I already created a repo. [https://github.com/akki-jat/style-loader-bug--332](https://github.com/akki-jat/style-loader-bug--332)

akki-jat:
@evilebottnawi @michael-ciniawsky any response?

michael-ciniawsky:
I see and I'm ok with using Object.keys here then so it at least throws, but I still think there should be a note added explaining this and inform devs that if their transform uses ES Modules they need to transpiled it beforehand

michael-ciniawsky:
Around the tranform option section something like

**`README.md#transform`**
```
> :warning: In case your transform uses ES Module Syntax you need to transpile it, or otherwise...
```

akki-jat:
@michael-ciniawsky you are right about the warning.
But what I am trying to say is right now even after the transpile it throws error.

### Why throws error even after the transpile?
export default is transpiled in exports.default and it returns the transform as object (not as a function) and actual transform function wrapped inside its default property.

So, meaning this will solve problem that is occurring after transpiling.

akki-jat:
@michael-ciniawsky any update?

akki-jat:
@evilebottnawi @michael-ciniawsky any response?

akki-jat:
@michael-ciniawsky @evilebottnawi please do something about this issue. I have to use my own forked library to overcome this problem.

akki-jat:
@evilebottnawi @michael-ciniawsky any response?

evilebottnawi:
@akki-jat why not traspile you code ?

akki-jat:
@evilebottnawi I am transpiling it. But what I am saying is default exports are traspiled different than normal export.
Take a look at pic.

image

lib/addStyles.js
```diff
@@ -254,7 +254,9 @@ function addStyle (obj, options) {
 
 	// If a transform function was defined, run it on the css
 	if (options.transform && obj.css) {
-	    result = options.transform(obj.css);
+	    result = typeof options.transform === 'function'
+		 ? options.transform(obj.css) 
+		 : options.transform[Object.keys(options.transform)[0]](obj.css);
```
michael-ciniawsky:
```diff
- options.transform[Object.keys(options.transform)[0]](obj.css)
+ options.transform.default(obj.css)
```

README.md
```diff
@@ -246,6 +246,8 @@ A `transform` is a function that can modify the css just before it is loaded int
 This function will be called on the css that is about to be loaded and the return value of the function will be loaded into the page instead of the original css.
 If the return value of the `transform` function is falsy, the css will not be loaded into the page at all.
 
+> Warning: In case you are using ES Module syntax in tranform.js then you have to transpile it otherwise it will throw error.
```
michael-ciniawsky:
```diff
- Warning:
+ :Warning:
```

```diff
- tranform.js
+ `transform.js`
```

```diff
- then you have to transpile it otherwise it will throw error.
+ , you **need to transpile** it or otherwise it will throw an `{Error}`.
```

akki-jat:
@michael-ciniawsky take a look now?

michael-ciniawsky:
Released in [`v0.23.1`](https://github.com/webpack-contrib/style-loader/releases/tag/v0.23.1)  🎉

PR #338, opened 2018-08-08
fix(addStyles): use `var` instead of `const` (IE fix)

pleunv:

v0.22.0 introduced support for webpack nonces, however the code contains a `const` keyword which throws on IE10 and below. This PR replaces `const` with `var` so everybody's happy. I've verified this fix in IE10.

### DIFF
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -209,7 +209,7 @@ function createStyleElement (options) {
 	}
 
 	if(options.attrs.nonce === undefined) {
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Released in [`v0.22.1`](https://github.com/webpack-contrib/style-loader/releases/tag/v0.22.1) 🎉

pleunv:
Awesome, thank you!

PR #341, opened 2018-08-24
feat(useable): add `insertInto` support (`options.insertInto`)

### LINKED ISSUE #327: style-loader/useable: passing insertInto through options
a while ago, the option to have insertInto as a function instead of a string was added to the style-loader. And this is exactly what I needed - and it works great. Then I tried using this with /useable, and I noticed that in index.js we have this:

if (typeof options.insertInto === "function") {
		insertInto = options.insertInto.toString();
	}
...
"options.insertInto = " + insertInto + ";",
"var update = require(" + loaderUtils.stringifyRequest(this, "!" + path.join(__dirname, "lib", "addStyles.js")) + ")(content, options);",

which makes the correct body of the function be included in the bundle,

while in the useable we have this:
"		dispose = require(" + loaderUtils.stringifyRequest(this, "!" + path.join(__dirname, "lib", "addStyles.js")) + ")(content, " + JSON.stringify(options) + ");",

which will omit the insertInto function.
Is there a particularity of the /useable that makes the usage with the insertInto (function) impossible?

jasonenglish:

feature. adds support for insertInto with `useable`. fixes #327 

Yes

I didn't, I figured this feature should be assumed to work with `useable` as well. I can add documentation to point it out though.

**Summary**

This feature is needed for injecting styles into the iframe with `useable` AFTER the page has loaded. Having the ability to call `style.use()` when the iframe is being setup is crucial. In this scenario we can't use the basic style-loader because specifying the insertInto option will attempt to look for the iframe when it doesn't exist yet.

No

**Other information**
n/a

### DIFF
--- a/test/useable.test.js
+++ b/test/useable.test.js
@@ -48,11 +48,11 @@ describe("useable tests", function () {
       requiredCssTwo = ".requiredTwo { color: cyan }",
       localScopedCss = ":local(.className) { background: red; }",
       localComposingCss = `
-      :local(.composingClass) {
-        composes: className from './localScoped.css';
-        color: blue;
-      }
-    `,
+        :local(.composingClass) {
+          composes: className from './localScoped.css';
+          color: blue;
+        }
+      `,
       requiredStyle = `<style type="text/css">${requiredCss}</style>`,
       existingStyle = `<style id="existing-style">.existing { color: yellow }</style>`,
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
Released in [`v0.23.0`](https://github.com/webpack-contrib/style-loader/releases/tag/v0.23.0) 🎉

Issue #320, opened 2018-04-24
Aborting CSS HMR due to changed css-modules locals when using react-hot-loader


pattiereaves:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using the latest version of style-loader and react-hot-loader, the page will reload every time a change is made to a css file (regardless of the change being to a property or a selector). It provides the following error when doing a full page reload: `Aborting CSS HMR due to changed css-modules locals.` I believe [this PR](https://github.com/webpack-contrib/style-loader/pull/298) introduced the issue.

**If the current behavior is a bug, please provide the steps to reproduce.**
I am not entirely sure what about my configuration is causing this issue, but I suspect it is the combination of using css modules, react hot loader and style loader. [See this comment](https://github.com/webpack-contrib/style-loader/pull/298#issuecomment-379346534). Downgrading to style-loader@0.19 fixes the issue. (Changes to CSS are applied to the page without a complete reload)

**What is the expected behavior?**
I can make changes to css and the changes will be applied using HMR.

**If this is a feature request, what is motivation or use case for changing the behavior?**
n/a

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**
node 8, npm 5
stylel-loader@0.21.0
"css-loader": "^0.28.11",
"react-hot-loader": "^4.0.1",
"webpack": "^4.2.0",
"webpack-dev-server": "^3.1.1"

tpolovyi:
I've faced the same issue while using `style-loader v0.21.0` alongside with `webpack-hot-middleware`.
[Here](https://github.com/tpolovyi/style-loader-demo) is a small demo repo where this bug is reproduced.

brigand:
@lydell this seems related to https://github.com/webpack-contrib/style-loader/pull/298 and there's a repro case linked here. 

It also works for me when adding a local on 0.19.1, and your change was published in 0.20. I'm using [babel-preset-react-hmre](https://www.npmjs.com/package/babel-preset-react-hmre).

asumaran:
I'm also getting the same error.

"style-loader": "^0.20.3",
"webpack-hot-middleware": "^2.22.2",

```
Error: Aborting CSS HMR due to changed css-modules locals.
    at main.js:157950
    at main.js:157953
    at Object.hotApply [as apply] (main.js:630)
    at cb (main.js:89783)
    at main.js:89799
onErrored @ main.js:89745
hotApply @ main.js:633
[...]

PR #347, opened 2018-10-03
Make HMR work when locals are changed, but with test repo to view it

KevinCChan:
Bugfix for https://github.com/webpack/style-loader/issues/320

No. There is not a direct way to test this change due to the many moving parts.

I created a repo at https://github.com/KevinCChan/css-hmr-bugfix to demo the change. All it does is open run a webserver on 127.0.0.1:8080 with a colored box to demo prefix and postfix behavior.

Yes

**Summary**
This is a bug fix for the crashing behavior introduced by https://github.com/webpack-contrib/style-loader/pull/298. There was a similar pull request, and we came to the same conclusion. I am doing my part and creating a test repository to acceptance test the changes so the fix can be integrated. I updated documentation as well.

No, I made this a toggle option specifically to avoid introducing a breaking change.

**Other information**
Please reach to me if there are any issues!

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ Styles are not added on `import/require()`, but instead on call to `use`/`ref`. 
 |**`singleton`**|`{Boolean}`|`undefined`|Reuses a single `<style></style>` element, instead of adding/removing individual elements for each required module.|
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`convertToAbsoluteUrls`**|`{Boolean}`|`false`|Converts relative URLs to absolute urls, when source maps are enabled|
+|**`manualReload`**|`{Boolean}`|`false`|Enable/Disable automatic full reload|
 
 ### `hmr`
 
--- a/README.md
+++ b/README.md
@@ -391,6 +391,20 @@ If convertToAbsoluteUrls and sourceMaps are both enabled, relative urls will be 
 }
 ```
 
+### `manualReload`
+
+[PR#298](https://github.com/webpack/style-loader/pull/298) introduced behavior for an automatic full reload if css-modules locals were changed by silently handling throwing an error. This works for most situations, but in certain situations - e.g. using webpack-dev-middleware directly in Express.js for remote/non-same host development, the thrown error will crash the loader. If manualReload is enabled, an error is not thrown and instead the console warns that css-modules locals has been changed. This restores the behavior to prior the introduction of PR#298 and resolves [this issue](https://github.com/webpack/style-loader/issues/320) where style-loader was aborting.
+
+**webpack.config.js**
+```js
+{
+  loader: 'style-loader',
+  options: {
+    manualReload: true
+  }
+}
+```
+
 <h2 align="center">Maintainers</h2>
 
 <table>
--- a/README.md
+++ b/README.md
@@ -393,7 +393,7 @@ If convertToAbsoluteUrls and sourceMaps are both enabled, relative urls will be 
 
 ### `manualReload`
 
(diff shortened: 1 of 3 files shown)


KevinCChan:
@evilebottnawi

There is no automatic test that can be written. I styled this PR similar to https://github.com/webpack-contrib/style-loader/pull/298 which was accepted without a test and provided a test repo to view the change.

evilebottnawi:
@KevinCChan don't think we need new option, we should fix it without new options

evilebottnawi:
Also this problem more related to dev server, not loader

KevinCChan:
@evilebottnawi

If we want to fix without new option, then I feel the correct course of action is to revert https://github.com/webpack-contrib/style-loader/pull/298.

https://github.com/webpack-contrib/style-loader/pull/298 specifically targets webpack-dev-server (a specific implementation of express/webpack-dev-middleware) at the cost of regressing functionality for every other implementation of express/webpack-dev-middleware.

I can look into alternatives more when I get back home tonight.

KevinCChan:
@evilebottnawi 

I did some digging around source code and playing with the repos and debuggers. This is what I found.

https://github.com/webpack-contrib/style-loader/pull/298 specifically targeted webpack-dev-server (WDS) with the hot option which meant that webpack-dev-server's included full reloaders would kick in.

If someone instead used WDS's hotOnly option in lieu of the hot option (which is documented [here](https://webpack.js.org/configuration/dev-server/#devserver-hotonly)), WDS behaves like a generic Express/webpack-dev-middleware/webpack-hot-middleware server configuration, much like my own and many others. You will see that https://github.com/webpack-contrib/style-loader/pull/298 actually fails in that specific configuration because the underlying assumption (of an error handler) no longer exists!

This makes the option toggle more relevant since it's impossible to determine at runtime whether or not an error handler has been provided to fully reload the page. Having the option at least allows users to configure the behavior of style-loader to match their given environment.

I recommend revisiting the test repo used as an acceptance test for https://github.com/webpack-contrib/style-loader/pull/298 (found [here](https://github.com/lydell/css-hmr-bug/tree/fix)) and editting the webpack_config from hot -> hotOnly and observe that issue https://github.com/webpack-contrib/style-loader/issues/320 existed then too.

sasha240100:
@KevinCChan @evilebottnawi Any updates to it?

lydell:
I think this is a good workaround, though I'm no expert (as evidenced by my previous PR, which btw, didn't intentionally target just webpack-dev-server but accidentally did so). :+1:

The "real", but much harder, solution is to add support in webpack's hot reloading "framework" for conditionally accepting hot updates depending on their contents, I think. Pinging @sokra just in case you have the time to share your wisdom, but I also understand if you're swamped with notifications and other stuff to do.

lydell:
Thinking a bit more about this, I thought that webpack-dev-server was just a convenience wrapper around webpack-dev-middleware, so to me it’s surprising that they behave differently. So maybe we should also make them behave the same? Not sure what the “correct” behavior is though.

KevinCChan:
@evilebottnawi @lydell 

This [issue](https://github.com/webpack/webpack/issues/418) led to the adoption of `hot` and `hotOnly` options in WDS.

TL;DR:
Sokra (webpack maintainer) and gaearon (react-hot-loader maintainer) discussed the exact issue that https://github.com/webpack-contrib/style-loader/pull/298 addressed about auto-reloading. It was decided that two implementations were necessary.

If the direction that style-loader wants to follow is one that is perfectly compatible with webpack then a new option is unavoidable due to the precedence set above in https://github.com/webpack/webpack/issues/418 and https://github.com/webpack/webpack/commit/e43a76a65e5511ebf58c9bc4983530869a48c921.

If style-loader only wants to support the `hot` option, then I will close this pull request.

KevinCChan:
@evilebottnawi No problem! I have been busy as of late so I did not really have time to respond. Let me see what is possible on the css-loader level

PR #364, opened 2019-01-26
hoist schema require, less work when empty options or hmr === false

hulkish:
This change aims to improve perf by reducing extraneous called to `validateOptions()` & hoisting `require()`.

perf

Yes

**Summary**

This change aims to improve perf by reducing extraneous called to `validateOptions()` & hoisting `require()`.

No

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,9 @@ var path = require("path");
 
 var loaderUtils = require("loader-utils");
 var validateOptions = require('schema-utils');
-var schema = require('./options.json')
+var schema = require('./options.json');
+
+var addStylesPath = path.join(__dirname, "lib", "addStyles.js");
 
 module.exports = function () {};
 
@@ -21,16 +23,17 @@ module.exports.pitch = function (request) {
 	var insertInto;
 
   if (options) {
-    validateOptions(schema, options, 'Style Loader');
+    validateOptions(schema, options, 'Style Loader (Useable)');
     options.hmr = typeof options.hmr === 'undefined' ? true : options.hmr;
 
-    if (typeof options.insertInto === "function") {
+    var insertIntoType = typeof options.insertInto;
+    if (insertIntoType === "function") {
       insertInto = options.insertInto.toString();
     }
 
     // We need to check if it a string, or variable will be "undefined"
     // and the loader crashes
-    if (typeof options.insertInto === "string") {
+    else if (insertIntoType === "string") {
       insertInto = '"' + options.insertInto + '"';
     }
   } else {
@@ -91,7 +94,7 @@ module.exports.pitch = function (request) {
 		"options.insertInto = " + insertInto + ";",
 		"",
 		// Add styles to the DOM
-		"var update = require(" + loaderUtils.stringifyRequest(this, "!" + path.join(__dirname, "lib", "addStyles.js")) + ")(content, options);",
+		"var update = require(" + loaderUtils.stringifyRequest(this, "!" + addStylesPath) + ")(content, options);",
 		"",
 		"if(content.locals) module.exports = content.locals;",
 		"",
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -3,18 +3,23 @@
...
 
 module.exports = function () {};
 
 module.exports.pitch = function (request) {
 	if (this.cacheable) this.cacheable();
 
-	var options = loaderUtils.getOptions(this) || {};
-
-	validateOptions(require('./options.json'), options, 'Style Loader')
+  var options = loaderUtils.getOptions(this);
```
evilebottnawi: Avoiding `|| {}` can break some places, it is good practice use `options.myOption`

index.js
```diff
@@ -3,18 +3,23 @@
 	Author Tobias Koppers @sokra
 */
 var path = require("path");
-
 var loaderUtils = require("loader-utils");
 var validateOptions = require('schema-utils');
+var schema = require('./options.json')
+
+var validOptions = new WeakSet();
```
evilebottnawi: Why?

hulkish:
```js
const Benchmark = require('benchmark');

const styleLoader = require('.');

Benchmark.options.minSamples = 100;

const loaderContextWithoutOptions = {
  cacheable() {},
  getOptions() {}
};

console.log('without options');
new Benchmark.Suite()
  .add({
    name: 'styleLoader.oldPitch()',
    fn() {
      styleLoader.oldPitch.call(loaderContextWithoutOptions, 'foo');
    },
  })
  .add({
    name: 'styleLoader.pitch()',
    fn() {
      styleLoader.pitch.call(loaderContextWithoutOptions, 'foo');
    },
  })
  .on('cycle', (event) => {
    console.log(String(event.target));
  })
  .on('complete', function() {
    console.log('Fastest is', this.filter('fastest').map('name'));
  })
  .run();

const options = {
  insertInto() { console.log('foo'); }
};
const loaderContextWithOptions = {
  cacheable() {},
  getOptions() { return options }
}

console.log('with options');
new Benchmark.Suite()
  .add({
    name: 'styleLoader.oldPitch()',
    fn() {
      styleLoader.oldPitch.call(loaderContextWithOptions, 'foo');
    },
  })
  .add({
    name: 'styleLoader.pitch()',
    fn() {
      styleLoader.pitch.call(loaderContextWithOptions, 'foo');
    },
  })
  .on('cycle', (event) => {
    console.log(String(event.target));
  })
  .on('complete', function() {
    console.log('Fastest is', this.filter('fastest').map('name'));
  })
  .run();
```

```sh
~/dev/repos/style-loader$ node benchmark.js
without options
styleLoader.oldPitch() x 37,352 ops/sec ±1.18% (184 runs sampled)
styleLoader.pitch() x 107,537 ops/sec ±1.36% (186 runs sampled)
Fastest is [ 'styleLoader.pitch()' ]
with options
styleLoader.oldPitch() x 39,174 ops/sec ±1.23% (185 runs sampled)
styleLoader.pitch() x 103,644 ops/sec ±1.27% (184 runs sampled)
Fastest is [ 'styleLoader.pitch()' ]
```

index.js
hulkish: invoking `require()` is not needed within `pitch`, it should happen once.

index.js
hulkish: This (I believe) matches behavior prior to my change. Since an empty options `{}` is subsequently populated (aka when expected props are `typeof options.x === 'undefined'`) - we can safely assume that our internal defaults are valid.

evilebottnawi: :+1:

evilebottnawi:
Better handle this case https://github.com/webpack/schema-utils

evilebottnawi:
We can discuss about this here https://github.com/webpack/schema-utils/issues/45

Issue #366, opened 2019-02-04
remove type="text/css" from inline css load


ovidijusr:
**What is the current behavior?**
when i use style-loader in webpack for inline loading I get inline style which looks like this:
```
<style type="text/css">
.css-code{}
</style>
```
**What is the expected behavior?**
ability to remove type attribute from style tag
```
<style>
.css-code{}
</style>
```

**If this is a feature request, what is motivation or use case for changing the behavior?**
I'm currently getting warnings from w3c about using this optional tag:

> Warning: The type attribute for the style element is not needed and should be omitted.
> From line 357, column 9; to line 357, column 31
> ↩</style><style type="text/css">/* sty

Tried to investigate why this type exist in first place but haven't found anything even compatibility related why.

Also tried to use attrs option for loader, but it doesn't seem to remove type attribute, and even if I try to disable it by using boolen it doesn't remove the attribute but makes it empty
```
attrs: {
  type: false,
},
```
Tried to remove it by giving empty string to attrs option but then it just returns empty attribute
```
<style type>
.css-code{}
</style>
```

Maybe i'm missing something? What is the reason it is being defined? Thank you for answers :)

evilebottnawi:
Feel free to send a PR

evilebottnawi:
Done in https://github.com/webpack-contrib/style-loader/pull/399

PR #367, opened 2019-02-04
Add ability to disable type attribute (#366)

ovidijusr:
Link to issue:
https://github.com/webpack-contrib/style-loader/issues/366#issue-406257824


ovidijusr:
Sorry for not setting the right user at first. I am the same user 😿

evilebottnawi:
@ovidijusr need fix it, do force push with right username

PR #368, opened 2019-02-04
Add option to remove "type" attribute #366

ovidijusr:
Issue: https://github.com/webpack-contrib/style-loader/issues/366


evilebottnawi:
@ovidijusr thanks, can you describe use case?

ovidijusr:
Well personally, I would set not having a "type" tag as default behaviour. Because [w3c markup validator](https://validator.w3.org/) gives warnings about that type. But this is breaking tests and i'm not sure if that won't affect something else.
Being said that, giving an option for this is a win/win because you can still use old behaviour and opt in for w3c correct one. [cssinjs had a similar case](https://github.com/cssinjs/jss/issues/724)

TLDR; if you use w3c - disabling type attribute helps to solve w3c warnings

evilebottnawi:
@ovidijusr what about remove this by default?

ovidijusr:
@evilebottnawi  Sure, should i update PR?

evilebottnawi:
@ovidijusr yep, let's do it

evilebottnawi:
Close in favor https://github.com/webpack-contrib/style-loader/pull/384

PR #371, opened 2019-03-22
Support only Node 6 or higher

hiroppy:

refactor, chore

no 

no(only automatic correction by prettier)

**Summary**

- Remove Mocha and Add Jest.
- Upgrade jsdom.
- Add husky, lint-staged.
- Add Prettier, eslint.

I refactor(e.g. `var` -> `const` or `let`) only tests not lib.

Maybe no because CI doesn't check Node0.12.0 from before.

**Other information**


evilebottnawi:
@hiroppy we use `webpack-default` (https://github.com/webpack-contrib/webpack-defaults) for loader/plugin and other stuff, it gives us contain identical configuration between projects, for using need update @webpack-contrib/defaults package and run npm run defaults

hiroppy:
@evilebottnawi You mean should we use webpack-default to replace existing code?

evilebottnawi:
@hiroppy it is boilerplate for us configuration, i can send pr to demonstrate that this

hiroppy:
Could you send pr to demonstrate that this?

evilebottnawi:
@hiroppy webpack-default already have prettier, we should use this boilerplate as standard for our projects to avoid difference configuration between our repos, and yes all new feature should be implemented in webpack-default and then updated in repos, i am open to new features, in Monday i will update webpack-default here, so you can see how i do this and do same in other repos

hiroppy:
@evilebottnawi ok, so I'll close this pr.

PR #372, opened 2019-03-25
Fix `variable is already defined` error caused in specific situation

key-moon:
If you bundle this module without minify, You can see these error : `i is already defined` ,`j is already defined` ,`domStyle is already defined`.

image

local variables are deleted when the function is completed, not block is completed.
So, This PR fix these multiple declaration.

Usually, loader is minified and packed in eval function so it doesn't matter,but in specific situation *1 that you can't minify code, this issue becomes tangible.

This PR doesn't contain any breaking changes as far as I know.

*1:  For example, to develop UserScript. Post minified UserScript is banned in some site.

### DIFF
--- a/lib/addStyles.js
+++ b/lib/addStyles.js
@@ -90,11 +90,13 @@ module.exports = function(list, options) {
 	addStylesToDom(styles, options);
 
 	return function update (newList) {
+        var i;
+        var domStyle;
 		var mayRemove = [];
 
-		for (var i = 0; i < styles.length; i++) {
+		for (i = 0; i < styles.length; i++) {
 			var item = styles[i];
-			var domStyle = stylesInDom[item.id];
+			domStyle = stylesInDom[item.id];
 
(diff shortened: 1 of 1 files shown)


evilebottnawi:
We need test or please create minimum reproducible test repo, thanks

key-moon:
[`here`](https://github.com/key-moon/style-loader-test/) is the minimum reproducible repo. The following link is the problem lines.
You can reproduce it by execute webpack.

https://github.com/key-moon/style-loader-test/blob/79991047bcbc0fcd0244d8c9eeb53f2b9f3b2b24/dist/bundle.js#L335-L336
https://github.com/key-moon/style-loader-test/blob/79991047bcbc0fcd0244d8c9eeb53f2b9f3b2b24/dist/bundle.js#L365

evilebottnawi:
Sorry, can't reproduce, all work fine

PR #375, opened 2019-03-25
refactor: webpack defaults

evilebottnawi:
webpack defaults

/cc @hiroppy friendly ping, here PR how we should update loaders/plugins


evilebottnawi:
Anyway after merge we need refactor code and tests

evilebottnawi:
Also we need fix coverage problem in test

hiroppy:
@evilebottnawi Thanks!

IMO: I think it would be difficult to track all plugins / loaders when the webpack-default's changes. I felt it would be easier to manage configuration files etc. collectively using monorepo like lerna.

evilebottnawi:
@hiroppy yep, anyway update webpack-defaults should be done on major release or if something buggy, in other cases we can avoid this update

hiroppy:
@evilebottnawi DO you plan to refactor other loaders/plugins using webpack-default?

evilebottnawi:
@hiroppy only when are we going to make major release

hiroppy:
ok, thanks

Issue #352, opened 2018-11-27
blob links seem applied to the dom after js execution occasionally


wusiquan:
hi, I found the sentence in README

> when style loader is used with { options: { sourceMap: true } } option, the CSS modules will be generated as Blobs...

then I enabled the `soruceMap` option, see the page in the latest Chrome(70.0.3538.110) when **developing**.

and js seems executed before the css applied to the dom `occasionally`, for instance, the page I write used `swiper` plugin, which calculate the `slideSize` would get the wrong size `occasionally` 
is it a chrome's bug? or the blob related problem?

evilebottnawi:
@wusiquan It can happen, you need load styles first when init swiper, use `import().then().catch()` for this. Looks `swiper` use `height` or `width` styles to calculate something

wusiquan:
@evilebottnawi Thank you for your replying. 
I tried the `import()` way, `Promise('./index.scss', './swiper.scss').all(() => new Swiper(...))`
but this time the `swiper` *always* calculate the wrong value...

After digging into the waterfall, I found the parse blob stylesheet section was hehind the execution
![waterfall](https://raw.githubusercontent.com/wusiquan/wusiquan.github.io/master/watherfall1-style-loader-issue352.png)

so, for temporoly solution, set the sourceMap option to false or init swiper in window.onload callback function, which let swiper plugin calculate the right value internally

wusiquan:
ok, please wait a while...I'll paste the repo link url

wusiquan:
all right, the [repo](https://github.com/wusiquan/style-loader-issue-352) has been created

evilebottnawi:
@wusiquan thanks!

wusiquan:
@evilebottnawi the same thanks to you. Hope it can help you reproduct.

rafaeldeoliveira:
this also happens in a project I'm working on

flut1:
I encountered this as well in a VueJS project. After vue has rendered the page (in the `mounted()` hook) I can see that the `<style>` tags have been inserted with blobs, but the styles are not applied to the page yet. If I wait a single tick (`setTimeout` of 0) the styles have been applied

Setting `sourceMap` to false also resolves the issue for me

4WakeStudio:
Same problem here :/.
Any better fix found?

flut1:
[...]

Issue #303, opened 2018-02-13
style-loader breaks source maps in Firefox 59


robertjk:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**What is the current behavior?**

When I have `style-loader` in my CSS loaders chain, the source maps do not work in Firefox. I see  error messages in the Developer Tools console like the one below. Setting `convertToAbsoluteUrls: true` doesn't change the situation.

`Source map error: Error: sourceMapURL could not be parsed
Resource URL: blob:http://localhost:8080/3a52944e-4117-4d8d-98d6-32cd525beae9
[...]

Issue #265, opened 2017-09-26
Does not apply objectUrl for `<link>` tag when setting `sourceMap: true`


Aqours:
**Move from**: [css-loader with `sourceMap: true` cause adding style tag delayed](https://github.com/webpack-contrib/css-loader/issues/613)

Actually, above issue is caused by style-loader. 

https://github.com/webpack-contrib/style-loader/blob/378e90637f1c2c30bff026d7f92bdc82d9d4ecb8/lib/addStyles.js#L354-L363

I have an idea to resolve this issue, but it's not perfect. Consider to apply this solution.

My solution code:
Using code `link.href = 'data:text/css;base64,' + btoa(encodeURIComponent(css));` replace line 363.

Test:
- Chrome 61: √
- Edge: √
- IE 11: √
- Firefox 55: × [Bug(maybe)](https://bugzilla.mozilla.org/show_bug.cgi?id=1403107)
- Safari: √

evilebottnawi:
@Aqours thank for issue, maybe best to wait when bug was fixed in `firefox` when change right source code?

Aqours:
@evilebottnawi can you provide a flag(e.g: `options.debug: true`) to enable this feature? so we can debug less/sass source code.

evilebottnawi:
@d3viant0ne what do you think about this issue?

WraithKenny:
I stumbled across this issue, not because of loading delays, but because the sourcemap doesn't actually work. In your tests, was the sourcemap actually loading, or just the stylesheet?

In my test, the stylesheet loads, but the styles in the firefox inspector point to the blob:null/ location, instead of the source file name. Additionally, it doesn't load the source file in the panel when you click on it.

This works correctly in Chrome. It looks like mozilla closed the referenced ticket as invalid. Maybe worth revisiting it?

wxlworkhard:
Why use `<link>` replace `<style>` when `sourceMap: true` ? `<link>` cannot load `url()` asset on Chrome. 

Why not just use
```
css += "\n/*# sourceMappingURL=data:application/json;base64," +
            btoa(unescape(encodeURIComponent(JSON.stringify(sourceMap)))) + " */";
``` 
and 

```
styleElement.appendChild(document.createTextNode(css));
```
in addStyles.js file. This is simple and have no problem.

PR #382, opened 2019-07-19
fix: inline style source maps

jonathantneal:
This PR contains a:

- [x] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Currently, if source maps are enabled, **style-loader** uses Blob URLs within `<link>` tags, instead of `<style>` tags with inline CSS. These Blobs URLs were originally used to workaround source map issues (apparently in earlier versions of Chrome) that are no longer present. 🤔

However, the use of `<link>` to load CSS is asynchronous, which can cause JavaScript to see an incorrect. This has caused numerous issues in projects. These issues were resolved by #219, but that fix was erroneously removed (likely due to human error during branch management). 🤷‍♂ 

There have continued to be issues with the Blob URL implementation. As a result, **Create React App** has removed the ability to use CSS source maps in development. 😢 

- **Bug Report**: https://github.com/facebook/create-react-app/issues/6399
- **"Fix" removing source map support**: https://github.com/facebook/create-react-app/pull/6472

Now I have work to do. And that work involves CSS. And when I work with CSS, I want source maps. I’ve done my research, discovered the preexisting work of others, and recreated the fix for you. 🙏 

If this is merged, the code will be shorter, faster, and ~peace~ source maps may be restored to the ~galaxy~ Create React App. 😂

### Breaking Changes

No, these are _unbreaking_ changes. 🤓

### Additional Info

This same fix was put into the [vue-style-loader fork](https://github.com/vuejs/vue-style-loader) in 2016. I would consider this a stable solution. 💪

- **Implementation**: https://github.com/vuejs/vue-style-loader/blob/master/lib/addStylesClient.js#L211
- **Original commit**: https://github.com/vuejs/vue-style-loader/commit/63f86710a096f3c5d117788ac83fdaabb80f762c


jonathantneal:
1. In order for this PR to pass, I had to run `npm audit --fix` and then `npm install --save-dev @commitlint/cli@8.1.0`. This failed because I performed it with a force push. I think you would have to reopen the PR.

2. Also, this PR incidentally fixes:
   - #265
   - #303
   - #352

Would you please reopen this PR?

PR #383, opened 2019-07-19
fix: inline style source maps

### LINKED ISSUE #265: Does not apply objectUrl for `<link>` tag when setting `sourceMap: true`

### LINKED ISSUE #303: style-loader breaks source maps in Firefox 59

### LINKED ISSUE #352: blob links seem applied to the dom after js execution occasionally

jonathantneal:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Currently, if source maps are enabled, **style-loader** uses Blob URLs within `<link>` tags, instead of `<style>` tags with inline CSS. These Blobs URLs were used to workaround [a bug that preventing source maps in `<style>` tag from working in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=456062), but that bug was fixed in 2015. 🤔

Unfortunately, using `<link>` to load CSS is asynchronous, which causes the FOUC. These issues were resolved by #219, which was approved, merged, and then “lost”; mostly likely due to a mistake in branch management. 🤷‍♂ 

There have continued to be issues with the Blob URL implementation. As a result, **Create React App** has removed the ability to use CSS source maps in development. 😢 

- **Bug Report**: https://github.com/facebook/create-react-app/issues/6399
- **"Fix" removing source map support**: https://github.com/facebook/create-react-app/pull/6472

But now I have work to do. And that work involves CSS. And when I work with CSS, I want to use source maps. So I’ve done my research, discovered the preexisting work of others, and restored the fix for you in this PR. 🙏 

If this is merged, the code will be shorter, faster, and ~peace~ source maps may be restored to ~the galaxy~ Create React App. 😂

### Breaking Changes

Yes, the breaking change is that the `convertToAbsoluteUrls` option is removed because it is no longer needed. That removal might be the reason [Codecov reports that this PR reduces coverage by 7.38%](https://github.com/webpack-contrib/style-loader/pull/383#issuecomment-513432778).

However, I would consider this an _unbreaking_ change. 🤓

### Additional Info

1. This same fix was already approved once and only removed by mistake, while it was correctly added to [vue-style-loader](https://github.com/vuejs/vue-style-loader) in 2016. That also means this is a stable fix. 💪

- **Implementation**: https://github.com/vuejs/vue-style-loader/blob/master/lib/addStylesClient.js#L211
- **Original commit**: https://github.com/vuejs/vue-style-loader/commit/63f86710a096f3c5d117788ac83fdaabb80f762c

2. In order for this PR to pass, I had to run `npm audit --fix` and then `npm install --save-dev @commitlint/cli@8.1.0`.

3. Also, this PR incidentally fixes these issues related to Blob:
   - Fixes #265
   - Fixes #303
   - Fixes #352


do-adams:
Awesome work @jonathantneal 

No more FOUC on development builds. That's really sweet, thanks 👍

PR #384, opened 2019-07-20
feat: remove type "text/css"

### LINKED ISSUE #366: remove type="text/css" from inline css load

jonathantneal:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

> there is very little reason to include this in modern web documents
— https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style#attr-type

Fixes #366

### Breaking Changes

No, this is not a breaking change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -693,16 +693,6 @@
         "regexpu-core": "^4.1.3"
       }
     },
-    "@babel/polyfill": {
-      "version": "7.4.0",
-      "resolved": "https://registry.npmjs.org/@babel/polyfill/-/polyfill-7.4.0.tgz",
-      "integrity": "sha512-bVsjsrtsDflIHp5I6caaAa2V25Kzn50HKPL6g3X0P0ni1ks+58cPB8Mz6AOKVuRPgaVdq/OwEUc/1vKqX+Mo4A==",
-      "dev": true,
-      "requires": {
-        "core-js": "^2.6.5",
-        "regenerator-runtime": "^0.13.2"
-      }
-    },
     "@babel/preset-env": {
       "version": "7.4.2",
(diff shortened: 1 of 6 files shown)


mistic:
@jonathantneal could you merge with master?

evilebottnawi:
@mistic it can be merged in near future, now we migrate tests

evilebottnawi:
@jonathantneal thanks for PR, closing in favor https://github.com/webpack-contrib/style-loader/pull/384

jonathantneal:
It is closed in favor of itself, @evilebottnawi ? 😝

evilebottnawi:
My mistake set-value :smile:

PR #452, opened 2019-12-28
notes about css-loader modules and proper configs in examples

quebits:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Make clear how to use `css-modules/locals`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ const divElement = document.createElement('div');
 divElement.className = styles['my-class'];
 ```
 
-All locals (class names) stored in imported object.
+All locals (class names) stored in imported object. To achieve this behaviour you also have to setup `modules` option for `css-loader`. For more information consult with `css-loader` [`documentation`](https://github.com/webpack-contrib/css-loader).
 
 **webpack.config.js**
 
@@ -118,7 +118,7 @@ module.exports = {
         use: [
           // The `injectType`  option can be avoided because it is default behaviour
           { loader: 'style-loader', options: { injectType: 'styleTag' } },
-          'css-loader',
+          { loader: 'css-loader', options: { modules: true } },
         ],
       },
     ],
@@ -177,7 +177,7 @@ module.exports = {
             loader: 'style-loader',
             options: { injectType: 'singletonStyleTag' },
           },
-          'css-loader',
+          { loader: 'css-loader', options: { modules: true } },
         ],
       },
     ],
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -315,7 +315,7 @@ module.exports = {
         test: /\.lazy\.css$/i,
         use: [
           { loader: 'style-loader', options: { injectType: 'lazyStyleTag' } },
-          'css-loader',
+          { loader: 'css-loader', options: { modules: true } },
```
evilebottnawi: We don't need `modules: true` in all examples, only where we say about modules

krail: All examples for `injectType` option contains accessing for css class names (via object itself or via `locals` property). As i can see, class names will be accessible only if `modules` set to `true`.

In case of 
```js
use: [
  { loader: 'style-loader' },
  { loader: 'css-loader' }
]
```

Result of

```js
import css1 from './1.css'
console.log(css1)
```

will be 
```js
{}
```

Another option value:
```js
use: [
  { loader: 'style-loader', options: { injectType: 'lazyStyleTag' } },
  { loader: 'css-loader' }
]
```

Result is:
```js
{ use: ƒ, unuse: ƒ }
```

There is no class names in object, nor locals property.

But if i set `modules: true` result will be:
```js
// 1st example, selector is a class name
{ selector: "tztrC7Uy9NarvcRmBTE5 _3-ZFXfyURwvBq_mDbTewjd" }

// 2nd example
{
  locals: { selector: "tztrC7Uy9NarvcRmBTE5 _3-ZFXfyURwvBq_mDbTewjd" }
  use: ƒ
  unuse: ƒ
}
```

krail: So i thought examples without explicit `modules: true` little bit confusing in parts with getting class names. Or may be there is a bug?

Anyway, i could remove `modules: true` for other `injectType` values.

evilebottnawi: Laze can be used without css modules

krail: Of course it will work, but there will be no `locals` if `modules` will be falsy.
This example has `divElement.className = styles.locals['my-class'];` which will not work without `modules: true`.

krail:
Possibly better approach would be total removing "locals" from "injectType" examples and create dedicated subsection about it?

evilebottnawi:
@krail i think yes

PR #478, opened 2020-08-13
fix: check if 'btoa' exists for old IE versions

okchangwon:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
If the `btoa` function does not exist, an error will occur as follows.

```
btoa is not defined
```


apple77y:
LGTM

PR #479, opened 2020-08-13
fix: check if 'btoa' exists for old IE versions

okchangwon:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
If the `btoa` function does not exist, an error will occur as follows.

```
btoa is not defined
```

### DIFF
--- a/src/runtime/injectStylesIntoStyleTag.js
+++ b/src/runtime/injectStylesIntoStyleTag.js
@@ -187,7 +187,7 @@ function applyToTag(style, options, obj) {
     style.removeAttribute('media');
   }
 
-  if (sourceMap && btoa) {
+  if (sourceMap && typeof btoa !== 'undefined') {
     css += `\n/*# sourceMappingURL=data:application/json;base64,${btoa(
       unescape(encodeURIComponent(JSON.stringify(sourceMap)))
     )} */`;


src/runtime/injectStylesIntoStyleTag.js
```diff
@@ -187,7 +187,7 @@ function applyToTag(style, options, obj) {
     style.removeAttribute('media');
   }
 
-  if (sourceMap && btoa) {
+  if (sourceMap && typeof btoa !== 'undefined') {
```
evilebottnawi, line 190: `btoa` should be `undefined` for old IE, can you provide screenshot?

okchangwon: image

okchangwon: 'typeof' should be used to check if global variables exist.

evilebottnawi: What is IE version? It is weird, because btoa should be false here

okchangwon: In IE9, `btoa` is `undefined`.
Or the following code is also good.
```javascript
if (sourceMap && window.btoa) {
```

evilebottnawi: I am fine with `typeof`, weird, I can't reproduce

okchangwon: Please check the console window on this page.

http://okchangwon.xyz/naver/ie9-btoa

image

evilebottnawi: error from `ie9-btoa` package, not from `style-loader`

okchangwon:
Let's recap the overall summary.

In the `applyToTag` method of `style-loader`
If `obj.sourceMap` option and `btoa` are present
I am adding `sourceMappingURL` to css.

The reason for checking the existence of `btoa` is
This is because there is no `btoa` function globally in Old IE(9).

Very good so far.

But the problem is,
The way to check for the existence of `btoa` is wrong.

It is currently as follows.
```javascript
 if (sourceMap && btoa)
```

If you access a nonexistent global variable (or function) like this
An error occurs regardless of browser type and version.

- IE11
> When accessing a variable that does not exist
image

- chrome
> When accessing a variable that does not exist
image

Because there is no btoa in 'IE9', an error occurs.
Even if 'btoa' exists in certain versions of 'IE9',
We recommend you to check in the right way in case you don't.

So I recommend the following method.

```javascript
 if (sourceMap && window.btoa)
// or
 if (sourceMap && typeof btoa !=='undefined')
```

Please let me know if I'm misunderstanding the situation.
Thank you.

PR #482, opened 2020-09-04
feat: added cssPostProcess option

rohm1:
This PR contains a:

- [ ] **bugfix**
- [x ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

When providing a theming option for customers, it is not practicable to build CSS assets for each and every single one of them, and that on a production server. Generating CSS with some placeholders and replacing them at runtime has proven to work nicely for us. The added runtime load is of course depending on how complex your want your runtime to be. In our case minimal.
We are currently working on prod with these patches and would now like to contribute the feature to the main stream.

### Breaking Changes

none

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,6 +32,10 @@ loaderApi.pitch = function loader(request) {
     insert: options.insert,
     base: options.base,
   };
+  const cssPostProcess =
+    typeof options.cssPostProcess !== 'function'
+      ? null
+      : JSON.stringify(options.cssPostProcess);
 
   switch (injectType) {
     case 'linkTag': {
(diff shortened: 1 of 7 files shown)


evilebottnawi (changes requested):
You should create loader for this, no new options here, it is runtime

rohm1:
@evilebottnawi thank you for the feedback. I cannot disclose more information as already written in the motivation, but I'm sure you can figure it out. I managed to build a loader that does the job and could remove the patch in our codebase.

PR #535, opened 2021-09-12
feat: allow to pass options to `insert` function through `style.use()`

### LINKED ISSUE #328: Allow importing code to determine where styles are inserted
This is a feature request.

### Problem

As far as I can tell, the style loader only allows for styles to be inserted into one location, either the `<head>` or where specified in the `inserInto` string or function.

This approach does not work with composed/nested custom elements, as each can have their own shadow dom. If, for example,  `insertInto` specifies the root custom element the styles are not picked up in children shadow doms.

### Feature Request
Allow importing code to determine where styles are inserted.

### Possible Solution
1. Add an `option` for local insertion. Let's call it `localInsert`, which defaults to `false` so that code executes as it currently does.
2. When `localInsert` is `true`, the `update` function is not invoked and we attach a method (`localInsertTo` ?) to `exports`, which allows for the importing file(s) to determine where the styles are inserted.

### Example
In the example below each custom element will have their imported styles inserted into their respective shadow doms by calling the `localInsertTo`.

`Webpack`
```js
...
options: {
  localInsert: true
}
...
```

`Parent`
```js
import css from './parent.css';

export class Parent extends HTMLElement {
  constructor() {
    super();
    this.shadow = this.attachShadow({ mode: 'open' });
    this.shadow.innerHTML = `
    <div>Parent 
      <x-child></x-child>
    </div>`;
    css.localInsertTo(this.shadow);
  }
}

customElements.define('x-app', Parent);
```

`Child`
```js
import css from './child.css';

export class Child extends HTMLElement {
  constructor() {
    super();
    this.shadow = this.attachShadow({ mode: 'open' });
    this.shadow.innerHTML = `<div>child</div>`;
    css.localInsertTo(this.shadow);
  }
}

customElements.define('x-child', Child);
```

OlegWock:
Allow to pass additional parameter to `style.use(anything)` which will be passed to `insert` function. This allows to insert each tag in different places and especially useful for Shadow DOM

Resolves #328

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I'm using style-loader in WebExtension and we heavily rely on Shadow DOM (we inject our widgets into 3rd party sites and want them look always nice, same on all sites). But it was a pain to inject styles into Shadow DOM (you better not peek on my old webpack config haha!). This small change allows to pass custom options (any object basically) to the `styles.use` call, which later will be passed to `insert` function. This allows customizing `insert` logic: now you can insert each style into different place and do whatever you want for each style individually

### Breaking Changes

This shouldn't break anything

### Additional Info

Resolves #328 (not in the proposed way tho)

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,6 +540,46 @@ module.exports = {
 
 Insert styles at top of `head` tag.
 
+You can pass any parameters to `style.use(anythingHere)` and this value will be passed to `insert` function.
+
+**webpack.config.js**
+
+```js
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.css$/i,
+        use: [
+          {
+            loader: "style-loader",
+            options: {
+              insert: function insertIntoTarget(element, options) {
+                var parent = options.target || document.querySelector("head");
+                parent.appendChild(element);
+              },
+            },
+          },
+          "css-loader",
+        ],
+      },
+    ],
+  },
+};
+```
+
+Insert styles to the provided element or to the `head` tag if target isn't provided. Now you can inject styles into Shadow DOM (or any other element).
+
+**component.js**
+
+```js
+import style from "./file.css";
+
+style.use({
+  target: document.querySelector('#myShadowDom').shadowRoot,
+})
+```
+
 ### `styleTagTransform`
 
 Type: `String | Function`
--- a/README.md
+++ b/README.md
@@ -540,7 +540,7 @@ module.exports = {
 
 Insert styles at top of `head` tag.
 
-You can pass any parameters to `style.use(anythingHere)` and this value will be passed to `insert` function.
+You can pass any parameters to `style.use(anythingHere)` and this value will be passed to `insert` function. These options will be passed to `styleTagTransform` function too. 
 
 **webpack.config.js**
 
--- a/src/index.js
+++ b/src/index.js
@@ -125,7 +125,8 @@ ${getInsertOptionCode(insertType, options)}
 options.domAPI = ${getdomAPI(isAuto)};
 options.insertStyleElement = insertStyleElement;
 
-exported.use = function() {
+exported.use = function(insertOptions) {
+  options.insertOptions = insertOptions || {};
   if (!(refs++)) {
     update = API(content, options);
   }
--- a/src/index.js
+++ b/src/index.js
@@ -126,7 +126,7 @@ options.domAPI = ${getdomAPI(isAuto)};
 options.insertStyleElement = insertStyleElement;
 
 exported.use = function(insertOptions) {
-  options.insertOptions = insertOptions || {};
+  options.insertOptions = insertOptions;
   if (!(refs++)) {
     update = API(content, options);
   }
--- a/src/runtime/insertStyleElement.js
+++ b/src/runtime/insertStyleElement.js
@@ -4,7 +4,7 @@ function insertStyleElement(options) {
 
   options.setAttributes(style, options.attributes);
 
-  options.insert(style);
+  options.insert(style, options.insertOptions);
 
   return style;
 }
--- a/src/runtime/styleDomAPI.js
+++ b/src/runtime/styleDomAPI.js
@@ -18,7 +18,7 @@ function apply(style, options, obj) {
 
   // For old IE
   /* istanbul ignore if  */
-  options.styleTagTransform(css, style);
+  options.styleTagTransform(css, style, options.insertOptions);
 }
 
 function removeStyleElement(style) {
--- a/test/fixtures/lazy-insertOptions.js
+++ b/test/fixtures/lazy-insertOptions.js
@@ -1,0 +1,7 @@
+import style from './style.css';
+
+style.use({
+  insertInto: document.body,
+  additionalStyles: '.some-element {color: red;}'
+});
+
--- a/test/lazyStyleTag-insertOptions.test.js
+++ b/test/lazyStyleTag-insertOptions.test.js
@@ -1,0 +1,53 @@
+/* eslint-env browser */
+
+import {
+  compile,
+  getCompiler,
+  getEntryByInjectType,
+  getErrors,
+  getWarnings,
+  runInJsDom,
+} from "./helpers/index";
+
+describe('lazyStyleTag insertOptions', () => {
+  it(`should pass "insertOption" to "insert" function`, async () => {
+    expect.assertions(3);
+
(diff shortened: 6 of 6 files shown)


alexander-akait:
Good idea, anyway can you provide examples:
> I'm using style-loader in WebExtension and we heavily rely on Shadow DOM (we inject our widgets into 3rd party sites and want them look always nice, same on all sites). But it was a pain to inject styles into Shadow DOM (you better not peek on my old webpack config haha!). This small change allows to pass custom options (any object basically) to the styles.use call, which later will be passed to insert function. This allows customizing insert logic: now you can insert each style into different place and do whatever you want for each style individually

I want to add more tests and examples here?

OlegWock:
@alexander-akait previously we injected style tag to all shadow doms present on page because there was no way to specify which component loaded styles:

**component.js**
```js
import css from './preferences.less';
import ReactDOM from "react-dom";
import React from "react";

class MyComponent extends React.Component {}

const attachMyComponentToPage = ($target, props) => {
    if (!$target) {
        $target = $(`<div id="MyComponentShadow" data-ad-view="true"></div>`);
        $('body').append($target);
    }
    let shadow = $target[0].attachShadow({mode: 'open'});

    ReactDOM.render(<MyComponent {...props} />, shadow, () => {
        css.use();
    });
};
```

**Settings for styleLoader**

```js
{
        injectType: 'lazyStyleTag',
        insert: function insert(el) {
            let id = '_' + Math.random().toString(36).substr(2, 9);
            console.log('Style loader insert:', el, id);
            el.setAttribute('data-ad-id', id);
            let shadows = document.querySelectorAll('[data-ad-view]');
            shadows.forEach(function (node) {
                if (node.shadowRoot) {
                    node.shadowRoot.appendChild(el.cloneNode(true));
                } else {
                    node.appendChild(el.cloneNode(true));
                }
            });
            if (shadows.length === 0) {
                document.querySelector('head').appendChild(el.cloneNode(true));
            }
        },
        styleTagTransform: function (css, style) {
            let id = style.getAttribute('data-ad-id');
            let styleNodes = [];

            let shadows = document.querySelectorAll('[data-ad-view]');
            shadows.forEach(function (node) {
                console.log('Handling node', node);
                if (node.shadowRoot) {
                    console.log('Found shadow, appending child');

                    let styleNode = node.shadowRoot.querySelector(`style[data-ad-id="${id}"]`);
                    if (styleNode) styleNodes.push(styleNode);
                } else {
                    console.log('Ordinary node, appending child');

                    let styleNode = node.querySelector(`style[data-ad-id="${id}"]`);
                    if (styleNode) styleNodes.push(styleNode);
                }
            });
            if (shadows.length === 0) {
                console.log('No data-ad-view elements found, appending to page head');
                let styleNode = document.querySelector(`style[data-ad-id="${id}"]`);
                if (styleNode) styleNodes.push(styleNode);
            }

            for (let styleNode of styleNodes) {
                if (styleNode.styleSheet) {
                    styleNode.styleSheet.cssText = css;
                } else {
                    while (styleNode.firstChild) {
                        styleNode.removeChild(styleNode.firstChild);
                    }
                    styleNode.appendChild(document.createTextNode(css));
                }
            }
        },
    };

```

Let me explain what all this mess does. I assume this isn't perfect solutions and there is way to do it better with current version (without proposed changes to .use method). Here, in `insert` function we assign random ID to the style tag and walk through all `[data-ad-view]` elements (it's our shadow dom containers) and append copy of `<styles>` into each shadow. Later, in `styleTagTransform` we retrieve this ID to find all `<styles>` tags we created on previous stage and populate them with css string.

Because we often display multiple widgets on same page and all widgets use multiple .css files (one for antd, one for fontawesome, one for custom styles) we end up with a bunch of `<styles>` tags in each shadow DOM (in addition to out hacky webpack config), while only small part of them are used in current Shadow DOM. 


With changes proposed in this PR you can pass parameters to `.use` which will be later passed to insert function. I have such use case in mind:

**component.js**
```js
import css from './preferences.less';
import ReactDOM from "react-dom";
import React from "react";

class MyComponent extends React.Component {}

const attachMyComponentToPage = ($target, props) => {
    if (!$target) {
        $target = $(`<div id="MyComponentShadow" data-ad-view="true"></div>`);
        $('body').append($target);
    }
    let shadow = $target[0].attachShadow({mode: 'open'});

    ReactDOM.render(<MyComponent {...props} />, shadow, () => {
        css.use({target: shadow}); // Note we're now passing shadow root to .use call
    });
};
```


**Settings for styleLoader**

```js
{
        injectType: 'lazyStyleTag',
        insert: function insert(el, options) {
            (options.target || document.querySelector('head')).appendChild(el);
        }
    };

```

I think there might be other use cases (maybe you need to know which component invoked particular styles) but this^^ is main one I had in mind. 

--- 

While writing this comment I thought it might be useful to pass options from `.use` call to `styleTagTransform` too. But I'd like to hear your opinion on all this

alexander-akait:
> While writing this comment I thought it might be useful to pass options from .use call to styleTagTransform too. But I'd like to hear your opinion on all this

sounds good too

Note - maybe help in future https://github.com/webpack-contrib/css-loader/pull/1368

OlegWock:
>  it might be useful to pass options from .use call to styleTagTransform too

Done. Also changed 
```js
options.insertOptions = insertOptions || {};
```

to just

```js
options.insertOptions = insertOptions;
```

in case someone would like to pass `false` or `null` here for whatever reason

alexander-akait (commented):
Let's add test for `unuse` to sure we remove the same target

OlegWock:
If I understand correctly, `insert` receives pre-made `<style>` tag. Reference to it stored internally and used to remove that tag (wherever it is). I.e. changes from this PR doesn't affect `.unuse()` in any way

alexander-akait:
Yep, just need to add test(s) to avoid break it/regression in future

OlegWock:
I'm not very proficient with Jest, I hope these tests are okay

alexander-akait:
Okay, I will finish tests after merge it

alexander-akait:
I will finish this PR

OlegWock:
@alexander-akait Thanks! Do you have any plans when to release new version (to npm) with these changes?

alexander-akait:
@OlegWock today/tomorrow, want to finish some features for css-loader, because it requires more logic here

PR #615, opened 2023-12-06
docs(security): Add warning

### LINKED ISSUE #613: Promotion of unsafe code
Documentation Is:

- [ ] Missing
- [X] Needed
- [X] Confusing
- [ ] Not Sure?

### Please Explain in Detail...

[Content Security Polcies](https://www.w3.org/TR/CSP2/) (CSPs) make the web a safer place. Authors can use unsafe policies, but those are clearly labelled so authors do not use them by accident. One such unsafe policy is `unsafe-inline`.

This loader heavily promotes unsafe-inline code: All available options for `injectType` except for `linkTag` are unsafe, including the default value (`styleTag`). This makes it a barrier for the adoption of tighter CSPs. As this loader is used on a lot of websites, this has a huge impact on the security of the web in general. 

As discussed in #306 and #487, CSPs allow to use nonces for inline code. However, these are not a proper solution. The [spec](https://www.w3.org/TR/CSP3/#security-considerations) is quite clear about the many drawbacks of nonces:

> Using a nonce to allow inline script or style is less secure than not using a nonce, as nonces override the restrictions in the directive in which they are present. An attacker who can gain access to the nonce can execute whatever script they like, whenever they like. That said, nonces provide a substantial improvement over 'unsafe-inline' when layering a content security policy on top of old code. When considering 'unsafe-inline', authors are encouraged to consider nonces (or hashes) instead.

### Your Proposal for Changes

- Change the default inject type to `linkTag`
- Clearly explain the security issues of all other inject types
- Clearly explain the security issues of nonces

xi:
This PR contains a:

- [X] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Fixes #613

### Breaking Changes

--

### Additional Info

I created these additions based on https://github.com/webpack-contrib/style-loader/issues/613#issuecomment-1843225551 by @alexander-akait. I hope this is what you had in mind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,10 @@ module.exports = {
 };
 ```
 
+## Security Warning
+
+This loader is primarily meant for development. The default settings are not safe for production environments. See the [recommended example configuration](#recommended) and the section on [nonces](#nonce) for details.
+
 ## Options
 
 - [**`injectType`**](#injecttype)
@@ -963,6 +967,8 @@ module.exports = {
 ```
 
 ### Nonce
+
+If you are using a [Content Security Policy](https://www.w3.org/TR/CSP3/) (CSP), the injected code will usually be blocked. A workaround is to use a nonce. Note, however, that using a nonce significantly reduces the protection provided by the CSP. You can read more about the security impact in [the specification](https://www.w3.org/TR/CSP3/#security-considerations). The better solution is not to use this loader in production.
 
 There are two ways to work with `nonce`:
 


evenstensberg:
Could you sign the CLA?

evenstensberg (commented):
Looks good. Will LGTM after CLA is signed

xi:
CLA is signed

evenstensberg (approved):
lgtm

xi:
I fixed the commit message

alexander-akait (approved):

